PR #51, opened 2024-10-19
feat: store previous tab setting when reloading in tab

### LINKED ISSUE #49: Remember track choice upon reload?
## Problem

Whenever the user reloads the page, or navigates back (for instance from a speaker page), the track choice is reset. I find this a bit annoying when checking out the speakers on tracks 2 or 3.

Recording: https://jam.dev/c/f917a798-4e67-474f-93fc-14a6b5a2e773

## Solution

If we'd like to implement this, I propose these possible solutions:

- Store track choice in sessionstorage
- Store track choice in url

By storing the track choice in sessionstorage, each new tab will have its own state, as long as the tab lives. The choice won't be sharable, but I don't really see the point in sharing the url with the state anyways. Both solutions solve the issue where you lose the state whenever you reload or go back and forth.

boyum:
Fixes #49

### DIFF
--- a/src/components/Speakers.tsx
+++ b/src/components/Speakers.tsx
@@ -9,6 +9,8 @@ import { ScheduleTrack } from '@/lib/schedule'
 import { Container } from '@/components/Container'
 import { DiamondIcon } from '@/components/DiamondIcon'
 
+const selectedTabKey = 'selectedSpeakerTab'
+
 function ImageClipPaths({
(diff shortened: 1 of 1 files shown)


Starefossen (approved):

boyum:
Not perfect, as there's a slight switch when the app is hydrated, but I guess it works?

PR #697, opened 2026-08-03
fix(sponsors): correct sponsor contact page info and default to accepted sponsors

Starefossen:
## Problem

`/admin/sponsors/contacts` listed **every** sponsor record for the conference — prospects, negotiations and lost deals alike — with no way to narrow the list, and several things it displayed were not backed by the data:

- Billing format rendered as "PDF via Email" whenever `invoiceFormat` was unset, presenting a guess as a recorded choice.
- EHF sponsors with no organisation number looked fully billable, even though EHF is addressed by org. number and cannot be delivered without one.
- The "Active sponsors" stat counted every record, `closed-lost` included.
- Saving the contact editor invalidated a tRPC query the table did not subscribe to (the rows came from the server component), so the table kept showing stale contact and billing data until a full reload.
- "Export Contacts" was a stub that `alert()`ed *"Export functionality would be implemented here"*.
- "Send Broadcast (N)" counted *sponsors that have contacts*, while the broadcast goes to every synced *contact address* — a different, larger number.
- The broadcast modal advertised a from-address of `Cloud Native Days <undefined>` when no `sponsorEmail` was configured, ignoring the server's own `sponsorEmail → contactEmail` fallback.

## What changed

**Accepted sponsors by default, with filters**

The page now opens on won deals and fetches rows through `sponsor.crm.list`, so filtering stays server-side (per `AGENTS.md`) and a saved edit refreshes the table it came from. Filters: stage (multi), billing state, contact state, and free-text search. `billingComplete` is a new optional filter on `SponsorCRMFilterSchema`; it is evaluated in the resolver rather than GROQ because the rule spans the sponsor document.

**Billing reported as recorded**

New `src/lib/sponsor-crm/billing.ts` is the single definition of "can this sponsor actually be invoiced":

| Rule | Why |
| --- | --- |
| Billing email required for **both** formats | Delivery address for PDF; documented fallback if EHF fails |
| Invoice format never assumed when unset | The schema default only applies to new records — older ones carry no format |
| EHF requires `sponsor.orgNumber` | EHF is addressed by organisation number; `contract-readiness` already treats it as required |

Gaps surface inline on the row (amber, with the full explanation on hover) and drive the "Ready to invoice" / "Incomplete billing" stats.

**Other accuracy fixes**

- Each sponsor now leads with its **primary contact**, badged as such — `contactPersons` order is incidental, so the row carrying the billing block could previously be an assistant rather than the person named on the contract.
[...]

PR #698, opened 2026-08-03
feat(sponsors): invoice worklist with readiness signal and CSV basis

Starefossen:
## Problem

The sponsor CRM's invoice view is the same kanban board, grouped by `invoiceStatus`. That answers *where each deal is* — but raising invoices is a batch job, and the board is built for moving one card at a time:

- **The billing details are not on the board.** Cards show name, tier, tags, amount and status. Billing email, EHF/PDF, PO reference and org. number live in the sponsor drawer or on the contacts page, so the job becomes: open every card, or work two screens at once.
- **No export.** Everything is retyped into the accounting system.
- **Blockers surface too late.** A sponsor set to EHF with no organisation number looks perfectly billable on the board; the problem shows up once someone is already keying it in. The `invoice` guard in the state machine only checks that a billing email exists.
- **Won deals with no contract value are hidden entirely** — `view: 'invoice'` filters on `contractValue != null`, so work nobody has priced is invisible to the person who needs to chase it.

## What this adds

`/admin/sponsors/invoices` — a worklist, one row per won deal, with everything an invoice needs side by side: amount + currency, tier and add-ons, invoice format, billing email, PO reference, billing comments, contract signed date, invoice status with sent/paid dates, and what (if anything) blocks it.

- **Opens on unfinished work** — not sent, sent, overdue. Paid and cancelled are one click away.
- **Server-side filters**: `invoiceStatus`, a new `invoiceReady` filter on `SponsorCRMFilterSchema`, and search.
- **`Export invoice basis`** writes the filtered set to CSV — plain numeric amounts, ISO dates, add-ons in one cell, and a `Blockers` column. Blocked rows are exported too: omitting them would read as "nothing to do here", which is the opposite of the truth.
- **`evaluateInvoiceReadiness`** = the billing rules from #697, plus an amount to bill and a signed contract. Deliberately stricter than the state machine's write-path guard, which cannot see the sponsor document on every path.
- **Amount provenance is explicit** — a tier-price fallback is labelled `From tier price` rather than passing as a negotiated value.
- **Org. number and billing email copy in one click** — the two fields that get mistyped into accounting systems.
- **Totals stay per-currency**, following the convention `lib/budget/income.ts` already sets: amounts in different currencies are never summed.

Deals with no amount are **listed and flagged**, not hidden. The board's own `contractValue != null` filter is left alone in this PR.

## Refactors carried along

- `calculateSponsorValue` moved to `lib/sponsor-crm/value.ts` (re-exported from the old path) so non-React callers stop importing from `components/`.
- Invoice status labels joined `lib/sponsor-crm/labels.ts`; `formatInvoiceStatusLabel` now reads from it.
- CSV escaping extracted to `lib/csv.ts` and shared with the contacts export instead of duplicated.

## Deliberately not in this PR

- **Tightening `INVOICE_GUARDS`** to the full readiness rules. It guards write paths, and some of them pass a synthetic partial state, so this needs its own change to avoid either false blocks or a fail-open guard.
- **Payment terms → computed overdue, invoice numbers, bulk stamping of `invoiceSentAt`.** These need a schema change and a migration.

## Verification

- `pnpm test` — 365 files, 4725 tests green (18 new: readiness rules, row derivation, per-currency totals, CSV shape).
- `pnpm typecheck`, `pnpm lint` (0 errors), `pnpm knip` clean.
- Visual inspection via `pnpm shoot` at 1440px and 393px, light and dark.

<h3>Greptile Summary</h3>

Adds a sponsor invoicing worklist for won deals.
- Introduces server-side invoice status, readiness, and search filtering.
- Adds invoice readiness evaluation, row derivation, per-currency totals, and CSV export.
- Adds responsive invoice table UI, navigation, loading state, and Storybook coverage.
- Extracts shared sponsor-value, invoice-label, and CSV utilities while preserving existing exports.

The PR appears safe to merge with no actionable correctness or security failures identified.

The invoice worklist remains conference-scoped, applies its filters server-side, receives the complete fields needed for readiness and export, and consistently derives displayed rows, totals, and CSV output from the filtered result set.

| Filename | Overview |
|----------|----------|
| src/components/admin/sponsor/SponsorInvoicesPageClient.tsx | Adds the filtered invoice worklist, summary totals, CSV download, and related navigation without an identified correctness issue. |
| src/components/admin/sponsor/SponsorInvoiceTable.tsx | Adds responsive invoice rows with billing details, readiness blockers, copy actions, and CRM links. |
| src/lib/sponsor-crm/invoice.ts | Centralizes readiness evaluation, invoice-row derivation, and currency-separated totals with coverage for principal states. |
| src/lib/sponsor-crm/invoice-csv.ts | Adds a one-row-per-deal invoice-basis export containing delivery, amount, status, date, and blocker fields. |
| src/lib/csv.ts | Extracts shared CSV escaping, formula neutralization, BOM generation, and filename construction while preserving contacts-export behavior. |
| src/server/routers/sponsor.ts | Adds server-side readiness filtering after the existing conference-scoped sponsor query. |
| src/server/schemas/sponsorForConference.ts | Extends the CRM filter schema with an optional boolean invoice-readiness selector. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart LR
  UI[Invoice worklist filters] --> API[sponsor.crm.list]
  API --> DB[Conference-scoped sponsor query]
  DB --> READY[evaluateInvoiceReadiness]
  READY --> ROW[toInvoiceRow]
  ROW --> TABLE[Responsive invoice table]
  ROW --> TOTALS[Per-currency totals]
  ROW --> CSV[Invoice basis CSV export]
```

<sub>Reviews (1): Last reviewed commit: ["feat(sponsors): invoice worklist with re..."](https://github.com/cloudnativebergen/website/commit/70c972961ce872c314a11a5d5839d846ad82c9c8) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=49649221)</sub>


Starefossen:
Went through all seven findings. Six fixed, one declined.

**Fixed**

- **Status array order (Copilot)** — selected statuses are now normalised to `STATUS_OPTIONS` order, so `{sent, not-sent}` and `{not-sent, sent}` are one query key rather than two.
- **`type="button"` on the copy control (Copilot)** — explicit, so a future `<form>` ancestor can't turn a copy into a submit.
- **Date formatting (Qodo)** — now uses `extractDateFromISO` from `lib/time.ts`. Worth noting *why* it isn't `formatDateSafe`: this column is read by an accounting system, and `14. mars 2026` isn't a date to anything but a human. Comment added so the next person doesn't "fix" it into a localised string.
- **Unescaped CSV header row (Qodo)** — headers now go through `csvRow` like data rows. Latent today (headers are static), real once the helper gets reused. Added `src/lib/csv.test.ts` covering escaping, formula-injection guarding and the header row.
- **Early blob revoke (Qodo)** — delayed revoke, matching the existing `downloadSvg` pattern.
- **Double list query (Qodo)** — the strongest finding. The second call existed only to size the `of Y` in the result line, and `sponsor.crm.list` projects the full expanded document (tier prices, add-ons, contacts, activity stats). Added a `sponsor.crm.count` procedure backed by a GROQ `count()`. It deliberately accepts **only** the GROQ-side status filter: readiness and billing are evaluated in the resolver over full documents, so a count that pretended to honour them would return a number for a different set than the caller asked for.

**Declined**

- **`getConferenceForCurrentDomain()` guard (Copilot)** — the premise doesn't hold. There is no `isUnknownHost` anywhere in `lib/conference/sanity.ts`, and the unknown-domain path sets `error = new Error('Conference not found for domain: ' + host)` before returning, so `conferenceError || !conference` does fire. It's also the guard every other admin route uses; changing it here alone would make this page inconsistent for no gain.

**Follow-up, not in this PR:** the contacts page (#697) has the same second-list-for-a-count pattern and can move to `crm.count` now that it exists.

Starefossen:
Both new findings were real and are fixed:

- **Bare `\r` in `csvCell` (Qodo)** — the regex only matched CRLF/LF. Now strips every CR/LF variant, so a pasted control character in a billing comment can't split one row into two. Test added.
- **`mailto:` parameter injection (Qodo)** — added `mailtoHref`, which links only a single unambiguous address and renders anything else as plain text. That kills the `?bcc=` class of injection and has the side benefit of making malformed stored addresses visible rather than clickable. Tests cover injection, lists, display-name form and junk.

Admin-facing and the data is organizer-entered, so the practical severity was low — but both fixes are a few lines and the CSV one is a correctness bug in a shared helper, so no reason to carry either.

PR #699, opened 2026-08-03
feat(invitations): admin-only issuing of visa invitation letters

Starefossen:
## What this is

Participants who need a visa ask organizers for a letter of invitation. Today that means writing one by hand in a word processor, every time, with the organizer's own idea of what a consulate wants to see.

This adds `/admin/invitations`: an organizer takes the details the applicant already sent them (email, form, whatever the request arrived in), fills them in, and gets a signed PDF — downloaded, emailed to the applicant, or both.

Scoped deliberately to **admin-only issuing**. There is no participant-facing request flow, so nothing here depends on identifying a ticket holder — which is the hard part, since attendees live in Checkin and have no local account.

## The privacy design, which is most of the point

Passport number, date of birth and nationality are **rendered into the PDF and then dropped**. They are:

- never written to Sanity — `recordInvitationLetter` has no parameter that could accept them,
- never logged — the error paths log the letter reference and the error, never the input,
- not stored inside a saved PDF either, because no PDF is saved.

What remains is an audit record answering *"did we issue a letter, to whom, and who stands behind it"*: name, email, role, reference, issuing organizer, timestamp.

`__tests__/api/trpc/invitation-letter.test.ts` locks this contract — it asserts the passport number appears in neither the audit write nor the logs, and pins the exact set of persisted keys, so a future change that starts storing them fails the build.

The trade-off: **re-issuing means re-entering the details.** The organizer holds them in the original correspondence, and the alternative is becoming the custodian of a passport register.

## What the letter contains

Grounded in what consulates actually check: applicant identity as written in the passport, the organizer's legal entity with org. number and address, event name/dates/venue, the applicant's role, registration reference, intended stay, an explicit statement that they will leave — and the cost line.

Cost coverage is stated **in both directions, every time**. Whatever the organizer does not tick is written out as borne by the applicant, because an omission reads as a claim: saying nothing about accommodation invites the assumption that the organizer covers it. `costCoverageSentence` is unit-tested across all four combinations.

The letter body is a pure, tested builder (`content.ts`) — the wording a consulate reads is covered by tests rather than buried in JSX.

## Caught by looking at the output

Rendering a sample letter surfaced a bug no test would have: the event dates came out as **"5.–6. november 2026"** — the Norwegian house format — inside otherwise English prose. Added `formatDateRangeLocalized` to `lib/time.ts` (Intl `formatRange`, so each locale collapses the range its own way) and used `en-GB` for letters. `formatDatesSafe` is unchanged.

Also: the closing paragraph and signature block are now kept together, so the signature is never stranded alone on a second page.

Page count: a minimal letter is one page; a fully-populated one (passport expiry, employer, stay dates, addressee, registration reference) runs to two, with the closing and signature together on the second.

## Verification

- `pnpm test` — 368 files, 4771 tests green (44 new: letter content, cost wording, the privacy contract, delivery and failure paths, date ranges).
- `pnpm typecheck`, `pnpm lint` (0 errors), `pnpm knip` clean.
- Admin page inspected via `pnpm shoot`; a real PDF rendered and read page by page.
- `/privacy` documents both halves — what is discarded and what is kept — per the repo rule on new data collection.

## Follow-ups, not here

- Participant-facing requests (needs Checkin identity, and the buyer is often not the attendee).
- Reusing the organizer signature captured for contracts — it lives in `localStorage` per browser, so it is not shared between organizers; the endpoint already accepts a `signatureDataUrl` when we want to wire it.
- An editable letter template, in line with getting organizers off Sanity Studio.

<h3>Greptile Summary</h3>

This PR adds an admin-only workflow for generating, downloading, emailing, and auditing visa invitation letters.
- Adds the invitation form, issue log, admin route, navigation entry, and Storybook coverage.
- Adds an admin-gated tRPC router, transient applicant-detail validation, PDF generation, and email delivery.
- Adds a minimal Sanity audit record that excludes passport details and scopes listings by conference.
- Adds localized date-range formatting, privacy-page disclosure, and unit/integration tests.

The date-validation defect should be fixed before merging because it can place a normalized wrong date in an official letter or silently omit an entered travel date.

The API accepts calendar-impossible date strings and independently optional stay endpoints; downstream formatting normalizes impossible values and emits a stay period only when both endpoints exist.

**Files Needing Attention:** src/server/schemas/invitationLetter.ts, src/lib/invitation-letter/content.ts, src/lib/time.ts

| Filename | Overview |
|----------|----------|
| src/server/routers/invitationLetter.ts | Adds organizer-gated issuance and listing, including PDF generation, optional email delivery, and minimal audit persistence. |
| src/server/schemas/invitationLetter.ts | Defines transient input validation, but date validation permits impossible dates and unmatched stay endpoints. |
| src/lib/invitation-letter/content.ts | Builds the English letter content and explicit cost wording, while omitting a lone arrival or departure date. |
| src/lib/invitation-letter/pdf.tsx | Renders the letter as an in-memory PDF with tables, letterhead, signature block, and fixed footer. |
| src/lib/invitation-letter/sanity.ts | Persists and lists conference-scoped audit records without passport, birth-date, or nationality fields. |
| src/components/admin/invitation/InvitationLettersPageClient.tsx | Adds the client issuance workflow, in-browser download, notifications, form clearing, and audit table. |
| src/lib/email/invitation-letter.ts | Sends the generated PDF as an attachment and correctly escapes all dynamic values placed in HTML text contexts. |
| src/lib/time.ts | Adds locale-aware date-range formatting used by the English invitation letter. |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
  participant Admin
  participant UI as Invitations Admin UI
  participant API as invitationLetter.issue
  participant PDF as PDF Renderer
  participant Email as Email Provider
  participant Sanity
  Admin->>UI: Enter applicant and passport details
  UI->>API: Submit transient details and delivery mode
  API->>PDF: Build and render invitation letter
  PDF-->>API: PDF buffer
  opt Email requested
    API->>Email: Send PDF attachment
    Email-->>API: Delivery result
  end
  API->>Sanity: Store non-passport audit record
  API-->>UI: PDF, reference, and delivery status
  UI-->>Admin: Download and/or confirmation
```

<sub>Reviews (1): Last reviewed commit: ["feat(invitations): admin page for issuin..."](https://github.com/cloudnativebergen/website/commit/652c79a146a368cba9ce8405a473b4688e58940f) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=49658578)</sub>


Starefossen:
Both parts of that finding were real — fixed and pushed.

**Impossible calendar dates.** `2026-02-30` matched the shape check, and `Date` rolls it forward to 2 March, so the letter would have printed a birth date the applicant never had onto a document a consulate checks against their passport. The schema now round-trips the components (`isRealCalendarDate`) and rejects it at the edge. A real leap day still passes — tested both ways.

**Half a period of stay.** Also real, and the same failure mode I designed the cost-coverage wording against: the content builder prints the stay only when it has both ends, so a lone arrival date was dropped in silence and the organizer would have believed they had declared something the letter never mentions. Now it is both or neither, with the error pointing at the departure field.

Three tests added.

Starefossen:
Worked through all 11. Nine fixed, one already handled, one partly wrong on its premise.

**The one that mattered most** — Copilot and Qodo found it from two angles: the client decided whether to download from `values.delivery`, mutable form state that the success handler then clears. For **Email only + a failed send** that meant the organizer got neither an email nor a file, while the toast claimed the PDF had downloaded. Since nothing is persisted, that cost the applicant a full re-entry of their passport details.

Fixed on both sides: the server withholds `pdfBase64` **only** when the email actually went out (Copilot's exposure point), and its presence in the response is now what drives the download — no dependence on form state at all. Tests cover both branches.

**Also fixed**
- `Field` labels wrap their control, so the association is real for screen readers and clicking the label focuses the input (Copilot).
- `a organizer` → `an organizer` (Qodo). Article now derives from the label; tested across all four roles.
- Logo warnings from this flow logged as `[contract-pdf]`. `rasterizeLogoToPngDataUrl` takes an optional `logTag`, defaulted so the contract path is unchanged (Qodo).
- `shadow-sm` → `shadow-xs` on the submit button (Qodo).
- Restored `components/admin/invitation/index.ts` and import through it (Qodo). I had deleted it earlier because knip flagged it as unused — wrong resolution; now it is used and knip is clean.
- Calendar-impossible dates and half a period of stay — already fixed in the previous round.

**Partly wrong: `formatRange` "can throw"**

It does not throw on a reversed range. Verified:

```
formatRange(6 Nov, 5 Nov) => "6 – 5 November 2026"
formatRange(NaN, …)       => RangeError: Invalid time value
```

The NaN case is the only throwing one, and it was already guarded before the call. But a reversed range printing `6 – 5 November 2026` on an official letter is bad output, so the ends are now ordered to always read forward. Dates stored backwards are a data problem to fix at the source, not to print.

368 files / 4780 tests green, lint and knip clean.

PR #703, opened 2026-08-03
feat(branding): reach every public route with the tenant theme, and default to Konf

Starefossen:
A conference could already set `theme.primaryColor` / `accentColor`, but the theme was injected in exactly one place — `Layout.tsx`, imported by three files. So `(stream)` and the public badge pages rendered in Cloud Native Days blue no matter what the tenant had configured.

## The seam

One `TenantThemeStyle` server component, rendered once per **public** route-group layout: `(main)`, `(workshop)`, `(stream)`, `(public)`, and `not-found`.

Deliberately **not** the root layout — that would theme the speaker portal and admin, which the owner explicitly ruled out: a speaker submitting to five conferences should get one familiar interface, and admin uses our palette as functional UI where a tenant colour could collide with colour that carries meaning. A shared parent route group would have been one seam instead of four, but it needs a large directory move *and still misses `not-found.tsx`*, which lives at app root and cannot be nested in a group.

`(main)`, `(workshop)` and `(stream)` already resolve the conference for their own reasons and pass it in — **no new query**. Only `(public)` resolves the host itself, through the `'use cache'` read that already carries both `domain:<host>` and `conferenceTag(id)`, so a branding save invalidates it.

**Fail-closed.** `getConferenceForDomain` returns a *truthy* `{} as Conference` for an unknown host, so a bare falsiness check would never fire — the same shape as the `scopedFetch` footgun and the fail-open tenant regression reverted in #696. This uses the canonical `isUnknownHost` guard and renders nothing, falling back to the house palette. Covered by a test.

`__tests__/app/tenant-theme-route-coverage.test.ts` asserts which groups inject the theme and which deliberately do not, matching *imports* rather than substrings — the excluded layouts carry a comment explaining why they have no injector, and a naive `.includes()` read that prose as a violation.

## Konf as the platform default

**`Logo.tsx` is deleted.** `ConferenceLogo` now generates the mark from the tenant's own name (`BrandWordmark` / `BrandMonogram`) as SVG with a viewBox, painted through the same `--brand-*` seam. Rationale: Cloud Native Days is a tenant like any other, and its wordmark belongs in Sanity, not in platform code. Previously any conference without an uploaded logo rendered CND's mark on their own site.

**⚠️ Needs confirming before merge** — see the question in the review thread.

**"Powered by Konf"** in the main site footer, linking to a new `PLATFORM_URL` (`NEXT_PUBLIC_PLATFORM_URL`, defaulting to `https://konf.app`). Deliberately not `platformBaseUrl()`, which resolves this app's own request host and *throws* in production when unconfigured — neither is acceptable in a footer on every page. It appears on the main site only: not on `(stream)` (a credit on a projected venue screen is noise), not on the public badge (platform marketing on a verifiable credential is wrong), not on admin or CFP. It is one self-contained component with one call site, so a plan entitlement can gate the render rather than hunting a scattered string.

**Brand-literal sweep:** ~35 user-facing fallbacks, more than the audit's estimate — email From names and signatures, Slack and web-push defaults, the PDF attestation footer, badge SVG and LinkedIn payloads, download filenames. Two were grep-evasive: `'Cloud Native Day'` singular and `'CLOUD NATIVE DAYS'` uppercase. Also `public/sw.js`, which had a hand-mirrored push title — now pinned to `PLATFORM_NAME` by the sw-source drift test — and the new-contract-template seed, which stamped one conference's name into every tenant's PDF header.

## Screenshots

`(stream)` and `(public)` rendered with real components under the real theme: default renders the calendar icon and "View Speaker Profile →" in blue, themed renders both purple. Same fixture, same layout, only the colour moves.

One bug the screenshots caught: a one-line generated wordmark for "Cloud Native Day Bergen" is about twice as wide as the logo it replaces — callers size by height and type has no fixed aspect ratio — which shoved the footer nav off its row. Fixed by wrapping past a max aspect ratio.

## Known, not fixed

`src/lib/pwa/default-mark.ts` still ships the CND logomark as the PWA icon fallback, so a tenant with no `logomarkBright` installs with Cloud Native Days' home-screen icon. resvg has no font configured, so a generated monogram would rasterise as blank glyphs; closing it means embedding a font or emitting initials as outlined paths. Documented in place.

Also: `scripts/shoot-story.mjs` now honours `SHOOT_PORT`. Without it a worktree agent silently screenshots whichever checkout owns `:6006`.

## Verification

`pnpm typecheck`, `pnpm lint` (0 errors), `pnpm format`, `pnpm knip`, full suite 375 files / 4853 tests.

<h3>Greptile Summary</h3>

Adds tenant-aware theming and replaces conference-specific platform fallbacks with Konf defaults.
- Injects tenant theme variables across public route groups while keeping admin and CFP surfaces platform-neutral.
- Generates fallback wordmarks and monograms from tenant names when uploaded logos are absent.
- Adds the “Powered by Konf” footer credit and centralizes platform naming, URLs, and filename slugs.
- Replaces Cloud Native Days literals across metadata, notifications, emails, badges, contracts, and admin tooling.
- Updates contract-template defaults and makes the Storybook screenshot port configurable.

Tenant resolution remains host-scoped and cache-tagged, unknown hosts fail closed, public routes receive the theme at their layout boundaries, and platform fallbacks consistently use Konf without altering authentication or tenant authorization boundaries.

| Filename | Overview |
|----------|----------|
| src/components/TenantThemeStyle.tsx | Introduces a guarded server-side seam that resolves or accepts the current conference and injects its validated theme. |
| src/components/BrandWordmark.tsx | Adds generated SVG wordmark and monogram fallbacks with tenant-aware colors and bounded aspect ratios. |
| src/components/ConferenceLogo.tsx | Replaces the hardcoded conference fallback logo with generated tenant or platform marks. |
| src/app/(public)/layout.tsx | Adds tenant-theme coverage to standalone public badge routes. |
| src/app/(stream)/layout.tsx | Reuses the stream layout’s resolved conference to theme venue-facing pages without another query. |
| src/lib/branding/platform.ts | Centralizes the Konf platform name, safe slug, and public marketing URL. |
| sanity/schemaTypes/contractTemplate.ts | Removes the tenant-specific contract header seed so application creation paths can supply resolved organizer branding. |
| public/sw.js | Changes the untitled push-notification fallback from Cloud Native Days to Konf and adds a source drift assertion. |

</details>

<sub>Reviews (1): Last reviewed commit: ["feat(branding): reach every public route..."](https://github.com/cloudnativebergen/website/commit/7a0586262c36d65e431e68f0356398efd7b9b2d0) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=49708843)</sub>

- Knowledge Base — [Push notifications](https://app.greptile.com/cloudnativedays/-/custom-context/knowledge-base/cloudnativebergen/website/-/docs/push-notifications.md)
</details>


Starefossen:
Two of the three findings were real and are fixed; the third I'm declining with reasoning.

**Unmocked clock in `PublicSurfaceTheming.stories.tsx`** — correct. `NextTalkDisplay` derives talk status from the current time, so the capture would drift as the fixture ages and churn Chromatic. Pinned with the house pattern (`beforeEach` swapping `globalThis.Date`, as in `Countdown.stories`).

**`h-full w-full` → `size-full`** — correct, and it's an explicit convention in AGENTS.md. Applied to all three occurrences in `MemeGenerator.tsx` rather than only the line this PR happened to move.

**Declining: "`shoot-story.mjs` spawns the package manager without the `rtk` prefix."**

Three reasons. The script already spawned the package manager exactly this way before this PR — the only change here was adding `SHOOT_PORT` support, so the finding is about pre-existing behaviour that this diff did not introduce. The `rtk` guidance in AGENTS.md addresses commands a developer or agent runs in a terminal, not a Node script's internal `spawn` of its own dev server. And `rtk` is not installed in this environment at all, so prefixing it would break the script outright — which is how I found this in the first place: the tool is aspirational in AGENTS.md rather than present.

Worth noting the `SHOOT_PORT` addition is load-bearing for review quality: without it, an agent working in a git worktree silently screenshots whichever checkout happens to own port 6006, and gets confidently wrong images of someone else's branch.

375 files / 4853 tests green, typecheck and lint clean.

Starefossen:
**Resolved the open question about the logo fallback — with evidence rather than a guess.**

The concern was that deleting `Logo.tsx` would switch the flagship Cloud Native Days site from its drawn wordmark to a generated one, if that conference has no `logoBright` uploaded in Sanity. The dataset is private, so an unauthenticated read returns nothing; instead I checked what the live site actually renders.

`https://2026.cloudnativedays.no/` contains:
- **5 elements with `aria-label="Conference logo"`** — the wrapper `ConferenceLogo` emits *only* on the `if (brightLogo)` branch
- **0 occurrences of `#3B82F6` / `#98CA3F`** — the gradient stops hardcoded in `Logo.tsx`

So the uploaded logo is set and in use, and the code fallback is not being rendered on that site today. Deleting it is safe.

Enqueued.

PR #704, opened 2026-08-03
fix(program): treat schedule dates as local days, not UTC midnight

Starefossen:
A live bug, found while auditing the platform's Norwegian assumptions and confirmed by reproduction.

## The bug

A schedule date is a `YYYY-MM-DD` string. `new Date('2026-11-05')` parses that as **UTC midnight**, which for any viewer west of UTC is the *previous* local day — in São Paulo (UTC−3) it is 4 November at 21:00. `stripTime` then strips to local midnight and yields the 4th.

Reproduced directly:

```
TZ=America/Sao_Paulo   10:00 local on 5 Nov 2026, schedule 2026-11-05 → isScheduleToday = false
TZ=Europe/Oslo         same instant                                    → isScheduleToday = true
```

So a conference anywhere in the Americas never lights up its "happening now" rail on the actual conference day — and `isScheduleInPast` marks the current day as already over. Invisible from Norway, which is presumably why it survived.

## The fix

Day comparisons here are deliberately against the **viewer's own clock** — an on-site attendee's "today" is their own, which is right — so the date string now has to be read the same way. `startOfLocalDay` builds the date from its parts rather than parsing the string, giving local midnight, with a fallback to the old path for a malformed value.

Applies to `isConferenceDay`, `isScheduleToday` and `isScheduleInPast`.

## Tests

Three regression tests that pin the viewer timezone rather than trusting the machine's, covering west of UTC, east of UTC, and a genuinely past day west of UTC (so the fix cannot simply make everything "today").

**Verified they catch the bug**: with the previous implementation restored, one fails; with the fix, all pass. A timezone regression test that passes either way is worse than none, so this was checked rather than assumed.

## Verification

`pnpm typecheck` clean, `pnpm lint` 0 errors, full suite 375 files / 4870 tests.

<h3>Greptile Summary</h3>

This PR corrects schedule-day comparisons to interpret `YYYY-MM-DD` values as local calendar days.
- Adds a local-day parser used by conference-day, today, and past-date checks.
- Adds timezone-specific regressions for viewers east and west of UTC.

The PR appears safe to merge, with the local-day behavior consistently applied and covered by timezone-specific regression tests.

The changed parser fixes the UTC-to-local day shift for valid schedule dates, and no concrete blocking or non-blocking defect remains in the reviewed paths.

| Filename | Overview |
|----------|----------|
| src/lib/program/time-utils.ts | Replaces UTC parsing of date-only schedule values with local calendar-date construction across all day-comparison helpers. |
| __tests__/lib/program/time-utils.test.ts | Adds isolated timezone regressions covering current and past schedule days in São Paulo and Oslo. |

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(program): treat schedule dates as lo..."](https://github.com/cloudnativebergen/website/commit/756572a986ef74210dc3d2a2d34eaa24ba957959) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=49719267)</sub>

### DIFF
--- a/__tests__/lib/program/time-utils.test.ts
+++ b/__tests__/lib/program/time-utils.test.ts
@@ -99,12 +99,16 @@ describe('program/time-utils.ts', () => {
  */
 describe('program/time-utils.ts — viewer timezone', () => {
   const withTimeZone = (tz: string, run: () => void) => {
+    const hadTz = 'TZ' in process.env
     const original = process.env.TZ
     process.env.TZ = tz
     try {
       run()
     } finally {
-      process.env.TZ = original
+      // Reassigning an undefined original would set the STRING "undefined" and
+      // leak a bogus zone into every later test in this worker.
+      if (hadTz) process.env.TZ = original
+      else delete process.env.TZ
     }
   }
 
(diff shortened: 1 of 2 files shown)


Starefossen:
All three findings were real and are fixed in `02ebad03`.

**1. Impossible dates silently rolled forward (Qodo).** `new Date(2026, 1, 30)` doesn't reject 30 February, it returns 2 March — so a malformed date compared as a real day instead of falling through to the fallback. The components are now round-tripped against the constructed date, the same check the invitation-letter date parsing uses.

**2. Trailing segments accepted (Copilot).** `split('-')` read `'2026-11-05-extra'` as valid. Replaced with an anchored regex.

**3. `process.env.TZ` restored to the string `'undefined'` (Qodo).** The helper captured `process.env.TZ` and reassigned it in `finally`; when TZ was originally unset that assigns the *string* `'undefined'`, leaking a bogus zone into every later test in the same worker. It now tracks whether the key existed and `delete`s it.

Tests for the first two added; 4872 tests green, typecheck clean.

PR #705, opened 2026-08-03
feat(migration): backfill identity for the three existing conference editions

Starefossen:
As the platform becomes tenant-configurable, the existing sites are at risk: **today the house defaults ARE Cloud Native Days' identity**, so those sites render correctly precisely because nothing is stored. Every step that neutralises a default toward Konf silently changes them — unless their identity becomes explicit data first.

This backfills that data while it is (almost entirely) a visual no-op.

## ⚠️ It also repairs a live regression

**#703 is already deployed** — confirmed from the service worker's `CACHE_VERSION = 'cndn-3175974f'`.

Before merging #703 I checked whether the deleted `Logo.tsx` fallback was in use, but I checked only `2026.cloudnativedays.no`, found an uploaded logo, and concluded it was safe. That was insufficient: `aria-label="Conference logo"` appears **5×** on Norway 2026 and **0×** on both Bergen sites. Both Bergen editions had no uploaded logo and were relying on the fallback, so they are now rendering a generated name-wordmark instead of the Cloud Native Bergen mark.

The Bergen logo writes here restore it, recovered mechanically from `git show 3175974f^:src/components/Logo.tsx`. The restoration was verified three ways: it round-trips byte-for-byte through `sanitizeSvgUpload` (the authoritative write gate); it rasterises via resvg to a PWA icon byte-identical to the current fallback at 512, 512-maskable and apple-touch; and the real live header and footer markup was re-rendered with the deployed CSS, injecting the logo exactly as `ConferenceLogo` + `InlineSvg` would, in light and dark — correct sizing, correct slate→white flip, no layout shift.

## What each edition actually renders

| | Norway 2026 | Bergen 2025 | Bergen 2024 |
|---|---|---|---|
| Theme stored | none | none | none |
| Logo uploaded | yes | **no** | **no** |
| Background | `cloud-native` | `cloud-native` | `cloud-native` |

Determined from the live sites, not inferred: no `:root{--brand-primary…}` block on any page and all three manifests serve the unthemed `theme_color: "#1D4ED8"`; the background pattern layer is byte-identical across all three over 60,584 characters with 98 icons at ~0.108 opacity — the dense `iconCount: 50` setting, not `subtle` or `none`.

## What it writes, only where absent

- `theme` = `{#1D4ED8, #06B6D4}`, matching the current defaults
- `backgroundPattern` = `cloud-native`
- `logoBright` + `logomarkBright` for **Bergen only**
- the eight `sponsorshipCustomization` prospectus strings, merged *under* existing keys — which is why Norway 2026's own `heroHeadline` survives

Values are hardcoded rather than imported from the constants they mirror: importing would mean that running this *after* neutralisation writes Konf's palette onto a Cloud Native Days site. A test asserts the literals still equal today's constants, as a tripwire.

Targeting uses the router's own `domainServesHost` predicate. Zero, two, or cross-matched targets abort before any patch is yielded.

## The one thing to decide before running

Storing a theme changes **four dark-mode declarations**, measured by rendering the deployed CSS with and without the exact `<style>` block `conferenceThemeCss()` emits:

| Utility | Mode | Today | Themed |
|---|---|---|---|
| `hover:bg-brand-cloud-blue-hover` | both | `#1E40AF` | ≈ `#1942B8` |
| `bg-brand-cloud-blue` | dark | `#1E40AF` | `#1D4ED8` |
| `bg-brand-gradient` | dark | `#1E3A8A`→`#155E75` | `#1D4ED8`→`#06B6D4` |
| `border-brand-cloud-blue` | dark | `#2563EB` | `#1D4ED8` |

The `.dark` rules read the same `--brand-primary`/`--brand-accent` but carry their own *darker* fallbacks, so a themed tenant loses the bespoke dark tints. That is the documented design, not a bug — but it is a decision, not something to slip in silently. Leaving the theme unset is worse: both modes drift completely the moment the palette moves.

## Deliberately NOT migrated

`homepageSections` is left alone for all three. None of them stores it, so the homepage renders `getDefaultSections()` — which is **phase-aware**. Writing the array would freeze Norway 2026 on "Featured Speakers" and stop it switching to "Program Highlights" when its schedule publishes. That is a behavioural regression, so the dry run prints a note per edition instead.

Also unreachable by any data backfill: two `<Sponsors>` call sites that render house copy with no props, and `src/lib/pwa/default-mark.ts`, which still ships the CND mark as the platform fallback for *new* tenants.

## Not run

This PR does not execute the migration. It runs through `.github/workflows/run-migration.yml`, which exports a dataset backup and dry-runs first. The README carries the full pre-merge / post-merge checklist.

`tsc`, `eslint`, `prettier`, `knip` clean; 376 files / 4896 tests. The migration also loads under the real `sanity migration run` runner.

<h3>Greptile Summary</h3>

The PR adds a guarded, additive Sanity migration that resolves three existing conference editions by routing domain and backfills their current visual identity without replacing stored values.
- Pins the existing theme, background pattern, and sponsorship copy for all three editions.
- Restores the legacy Bergen logo and logomark for the 2024 and 2025 Bergen editions.
- Aborts when targets are missing, ambiguous, or cross-matched, and skips drafts.
- Adds migration documentation and tests for targeting, value preservation, idempotency, SVG sanitization, and rendered PWA icon equivalence.

The PR appears safe to merge, with no concrete changed-code defect identified in the guarded backfill.

The migration resolves all targets before producing changes, preserves existing configured values, skips drafts, and has focused coverage for targeting, idempotency, and restored brand assets.

| Filename | Overview |
|----------|----------|
| migrations/046-conference-identity-backfill/index.ts | Defines the migration I/O, validates target resolution before yielding changes, skips drafts, applies planned additive patches, and checks that every resolved target appeared in the stream. |
| migrations/046-conference-identity-backfill/plan.ts | Encapsulates domain-based targeting and field-level backfill planning while preserving existing themes, branding assets, sponsorship values, and phase-aware homepage configuration. |
| migrations/046-conference-identity-backfill/plan.test.ts | Thoroughly tests target resolution, hardcoded fallback parity, preservation behavior, idempotency, SVG sanitization, and icon rendering equivalence. |
| migrations/046-conference-identity-backfill/legacy-brand.ts | Stores the recovered legacy Bergen wordmark and logomark as sanitized inline SVG constants for the two affected editions. |
| migrations/046-conference-identity-backfill/README.md | Documents the migration rationale, operational safeguards, known visual deltas, execution checklist, verification steps, and rollback procedure. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A[Fetch published conferences] --> B[Resolve three target hosts]
    B --> C{All targets unique?}
    C -- No --> D[Abort before yielding patches]
    C -- Yes --> E[Stream conference documents]
    E --> F{Resolved target?}
    F -- No --> E
    F -- Yes --> G[Plan absent identity fields]
    G --> H{Any fields to set?}
    H -- Yes --> I[Yield additive patch]
    H -- No --> J[Mark target seen]
    I --> J
    J --> E
    E --> K{Every target seen?}
    K -- No --> L[Report incomplete stream]
    K -- Yes --> M[Print migration summary]
```

<sub>Reviews (1): Last reviewed commit: ["feat(migration): backfill the three exis..."](https://github.com/cloudnativebergen/website/commit/d120030e94f928d31cde911ad71cc9cd9f427277) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=49725144)</sub>


Starefossen:
Both findings were real and are fixed in `2b038243`.

**1. `isBlank()` treated any non-string as blank.** This one mattered: it broke the single promise the migration makes. A `theme.primaryColor` stored as a number, or a `logoBright` stored as an image reference by some later schema revision, read as *absent* and would have been overwritten with the Cloud Native default — silently, on exactly the documents where a wrong value is hardest to spot. `isBlank` is now null/undefined/whitespace-only-string, nothing else; wrong-typed data is schema validation's problem, not a backfill's. Locked with a test I verified fails against the old predicate.

**2. The pre-pass overfetched.** The targeting query projected the identity fields — including two ~19KB inline SVGs — for every conference in the dataset, while only ever comparing `domains[]`; every field check runs against the streamed document, not that row. Now projects `{ _id, title, domains }`, and `resolveTargets` takes a narrower `ConferenceTargetDoc` so the projection can't quietly widen again.

On the `rtk` prefix suggestion in the README: declining, same as on #703 — `rtk` is not installed in this environment at all, so prefixing the command would break the documented invocation rather than fix it.

PR #706, opened 2026-08-03
feat(homepage): define every section's rendering in every lifecycle state

Starefossen:
64% of 2026 Kubernetes Community Day events have no website at all. Konf's competition for those organisers is nothing — so what decides adoption is whether a brand-new event with no speakers, no sponsors, no photos and no schedule gets a homepage that looks *deliberate* on day one.

The canonical failure is live on another conference's site today: it renders `0` for all four of its auto-computed stats rather than hiding the section.

**We had the same bug.** Verified by reverting these files in a running Storybook and shooting the regression story — it reproduced exactly: `Sessions 0+ / Speakers 0+ / Workshops 0 / Days 1 / Topics 0+ / Tracks 0`, plus "1 days", plus a "Get Your Tickets Now" button under "Don't Miss These Standouts" with no standouts beneath it.

## The state model

One ordered **stage** plus orthogonal **facets**, because a flat enum would have to be the cross product.

- **Stage**: `announced` → `cfp-open` → `curating` → `programme` → `post-event`, plus `cancelled` / `archived`
- **Facets**: cfp, tickets, content flags, and a resolved primary CTA

Everything except `cancelled` and `archived` is **derived** from fields organisers already fill in — an organiser should not have to flip switches to get a sensible page.

Two signals the code was throwing away:

- **Sold out / not yet on sale** was unknowable because `page.tsx` discarded `ticketData.tickets` and kept only the price. Now derived from each type's sale window plus availability — and it **fails closed on the claim**: `sold-out` requires every active public type to positively report zero, and a single `null` degrades to `on-sale`. A wrong "sold out" costs an organiser real money.
- **`hasRecordings`** from the existing attachment model, so "watch the talks" is only promised when a recording actually exists.

`isFirstEdition` is a **documented proxy, not a fact** — there is no cross-edition link in the data model at all, so it uses "no past photos and no vanity metrics", which answers the only question the page asks: *is there history to show?*

## Per section

| Section | Empty behaviour | Why |
|---|---|---|
| ProgramHighlights | Only tiles with a non-zero number; `null` when the schedule holds no confirmed talk | A zero is never a fact about a conference, only about data not entered |
| SaveTheDate (new) | Deliberate placeholder | The one place a placeholder beats hiding — day one has nothing else |
| Gallery / Metrics / FAQ / Venue / Countdown | Hide | A heading over nothing is worse than nothing |
| Sponsors | Logos render; the "Become a Sponsor" pitch is suppressed post-event | After the fact it asks for money for something that already happened |
| Hero | Programme button gated on programme *content*, not the publish date; no ticket button when sold out or not on sale | A "View Program" that opens an empty page is the same broken promise as the zero band |
| Cancelled / Archived | **Replace the page** | Visitors act on buttons, not paragraphs |

## Four bugs found by looking, not by reading

1. **Day one** rendered the countdown as `200 00 00 00` — three-digit day counts overflowed the column at 393px, which is the *normal* case for a save-the-date.
2. **Day one** made "Become a Sponsor" the loudest, first, green button on a brand-new event's homepage. The visitor-facing link now leads.
3. **Roadmap rows** used the action button as status text — "Open until 15. apr. 2026" wrapping to two lines. Split into a status line and an action.
4. **Archived** printed the last edition's dates under a tombstone, reading as though it were still being scheduled. Dates now appear only on `cancelled`, where "was going to happen on these dates" is the point.

## Zero migration, proven

`HomepageComposition` **Default and CustomDark are SHA-256 byte-identical** before and after at 1280×4200, against main's own story fixture. `getDefaultSections` has exactly two departures from the old layout, and both fire only where the page previously rendered a hole.

Cancelling is also now possible **without Sanity Studio** — a `lifecycle` fieldset, its own tRPC mutation and a settings card.

## Also fixed

Event JSON-LD now reports `EventCancelled`, emits no Offer for cancelled or archived events, and reports `SoldOut` for a sold-out sale. The file already stated the principle — that structured data must never claim tickets are purchasable while the site hides every ticket CTA — it just predated these states.

## Known, not fixed

- The house locale is `nb-NO`, so a Trondheim save-the-date renders "17. september 2026". Out of scope here; it is the next thing a non-Norwegian organiser notices.
- Only the homepage is lifecycle-aware — a direct link to `/tickets` on a cancelled event still shows a checkout.

376 files / tests green; typecheck, lint, format clean.

<h3>Greptile Summary</h3>

The PR introduces a derived homepage lifecycle model and applies it across homepage composition, ticket availability, structured event data, and administration.
- Adds lifecycle-aware rendering, save-the-date and terminal-state notices, and content-sensitive homepage sections and CTAs.
- Derives public ticket availability from provider inventory and propagates it to homepage and JSON-LD output.
- Adds organizer controls for cancelling, archiving, and restoring an event.
- Expands lifecycle, section-rendering, ticketing, SEO, and Storybook coverage.

The empty-programme composition defect should be fixed before merging because it removes the deliberate save-the-date experience from a realistic upcoming-event state.

A passed programme publication date changes the stage to programme even without confirmed talks, causing the default section selector to suppress Save the Date while selecting no content-bearing replacement.

**Files Needing Attention:** src/lib/homepage/sections.ts and src/lib/homepage/lifecycle.ts

| Filename | Overview |
|----------|----------|
| src/lib/homepage/lifecycle.ts | Introduces the central lifecycle stage and facet derivation; its date-only programme stage can diverge from actual programme content. |
| src/lib/homepage/sections.ts | Defines lifecycle-aware default composition, but drops Save the Date when a programme date has passed without programme content. |
| src/components/homepage/SectionRenderer.tsx | Applies lifecycle state to section visibility, CTAs, sold-out messaging, and terminal-state page replacement. |
| src/lib/tickets/public.ts | Adds fail-closed derivation of upcoming, on-sale, sold-out, and unknown ticket availability. |
| src/server/routers/conference.ts | Adds a tenant-resolved, schema-validated mutation for lifecycle status and notice fields. |
| src/lib/seo/eventJsonLd.ts | Updates event status and offer output for cancelled, archived, and sold-out lifecycle states. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
  A[Conference fields and current time] --> B{Cancelled or archived?}
  B -- Yes --> C[Replace homepage with lifecycle notice]
  B -- No --> D[Derive CFP, ticket, and content facets]
  D --> E[Resolve lifecycle stage]
  E --> F[Choose default homepage sections]
  F --> G[Render content-aware sections and CTAs]
  D --> H[Generate ticket-aware Event JSON-LD]
```

<sub>Reviews (1): Last reviewed commit: ["fix(seo): stop advertising a cancelled e..."](https://github.com/cloudnativebergen/website/commit/42b113ec4976f8b9037d68d6d0e772fca68acb06) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=49726525)</sub>

### DIFF
--- a/sanity/schemaTypes/conference.ts
+++ b/sanity/schemaTypes/conference.ts
@@ -1357,7 +1357,7 @@ export default defineType({
             type: 'text',
             rows: 2,
             description:
-              'Optional copy under the heading. Defaults to the dates and city.',
+              'Optional extra copy. There is no default: the card already shows the dates, the venue and city, a countdown and the milestone list, so leaving this empty simply adds no extra line.',
           }),
         ]),
         defineHomepageSection('homepageFeaturedSpeakers', 'Featured Speakers', [
--- a/sanity/schemaTypes/conference.ts
+++ b/sanity/schemaTypes/conference.ts
@@ -144,6 +144,15 @@ export default defineType({
         'Ordered list of homepage sections. Leave empty to render the default phase-aware layout (hero, gallery, featured speakers / program, sponsors).',
       options: { collapsible: true, collapsed: true },
     },
+    // Homepage lifecycle OVERRIDE (F5). Its own fieldset, appended last, for the
+    // same merge-conflict reason as `homepage` above.
+    {
+      name: 'lifecycle',
+      title: 'Event Status',
+      description:
+        'Only for calling an edition off or retiring it for good. Every other state (save-the-date, CFP open, programme published, post-event) is derived from the dates above and needs no switch here.',
+      options: { collapsible: true, collapsed: true },
+    },
   ],
   fields: [
     // === Basic Information ===
@@ -1335,6 +1344,22 @@ export default defineType({
             ],
           }),
         ]),
+        defineHomepageSection('homepageSaveTheDate', 'Save the Date', [
+          defineField({
+            name: 'heading',
+            title: 'Heading',
+            type: 'string',
+            description: 'Optional heading. Defaults to "Save the date".',
+          }),
+          defineField({
+            name: 'description',
+            title: 'Description',
+            type: 'text',
+            rows: 2,
+            description:
+              'Optional copy under the heading. Defaults to the dates and city.',
+          }),
+        ]),
         defineHomepageSection('homepageFeaturedSpeakers', 'Featured Speakers', [
           defineField({
             name: 'heading',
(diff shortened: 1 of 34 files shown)


Starefossen:
All nine findings assessed in `c7a18098` — eight fixed, one declined.

**The P1 (greptile + qodo, independently).** Real and the most serious. `getDefaultSections()` offered Save-the-Date only for `announced`/`cfp-open`/`curating`. Once `programDate` passed, the stage became `programme` — but with no confirmed talks and no featured speakers, `hasProgramme` is false, so the programme content section didn't render *either*. A live homepage went near-empty because a date rolled past, with no organizer action involved.

The stage gate now covers the whole pre-event span while the content gate is unchanged, so the band appears exactly while there's nothing else to lead with. Evidence this was the original intent rather than a judgement call: `resolveRoadmapSteps` already has a branch for "programme date passed, nothing published", and that branch is reachable *only* from inside the Save-the-Date band — the roadmap assumed the band renders in this state all along; only `getDefaultSections` disagreed.

**Programme CTA tracked as `events.info` (copilot + qodo).** Real. It has its own registered events now, following the `cta-<intent>-<location>` scheme already in `PIRSCH_EVENTS`.

**"Watch the talks" shown pre-event (qodo).** Real. `hasRecordings` only requires a recording URL on a confirmed talk, which can exist before the event, so the label is now gated on `post-event` as well.

**"Defaults to the dates and city" (copilot, twice).** Real — but I fixed the copy, not the code. The card already leads with the date range as its `<h2>` and the venue and city on the line directly below, so a derived default would restate both a third time. The promise was the stale part. A test now asserts the date and city each appear exactly **once**, so it fails if anyone adds the default later.

**Mocking internal components (qodo).** Real. `SectionRenderer.test.tsx` mocked `SaveTheDate` and `LifecycleNotice`, so it asserted against stand-ins. Removing them exposed a second problem the reviewers didn't catch: the `next/link` mock dropped every prop, which had quietly made any assertion on `href` or `data-pirsch-event` vacuous. `SaveTheDate` also had no story and no test at all — both added.

**Declined: "`pinClock()` is never restored" (qodo).** Not a real leak. `pinClock` returns a restore closure (`lifecycle.mocks.ts:46`) and `beforeEach: () => pinClock()` is a concise-body arrow, so Storybook already receives it — the same shape `Countdown.stories.tsx` uses. I added a comment at the call site naming the mechanism so it doesn't get re-raised.

Also fixes the knip failure that was blocking Quality Checks: `export const soldOutConference = midCycleConference` made two exports the same object.

4953 tests green, typecheck clean, knip clean.

PR #707, opened 2026-08-03
feat(homepage): a sanitised rich-content block, and harden the existing one

Starefossen:
Research across 20+ real conference homepages found that nearly every distinctive event has exactly **one weird thing** no fixed section vocabulary can express: Cloud Native Days Vienna renders its venue as a Kubernetes YAML manifest; ZurichJS embeds a live social feed; Rails World crowdsources venue suggestions; JSConf México keeps a wall of 30+ non-paying community logos deliberately separate from sponsors.

The section registry is deliberately closed — no raw HTML, no embeds — for XSS and brand safety on a shared multi-tenant deployment. The decision taken here is a **sanitised** escape hatch, not an open one.

## It extends the existing block rather than adding a new one

`homepageRichText` already *was* "heading + portable text". A second block would have duplicated the schema, the Zod arm, the router case, the renderer, the label map and the editor panel — and organisers would face two indistinguishable entries in the "Add section" menu.

More importantly: the existing block's `content` was validated by `z.object({_type: z.string()}).catchall(z.unknown())` — **a passthrough**. A hardened new block sitting next to an unhardened old one would have left the actual hole open. Extending closes it and adds the capability in one move.

## The allowlist

| Type | Fields |
|---|---|
| `block` | style ∈ normal/h1–h4/blockquote · listItem ∈ bullet/number · level 1–5 · children (spans only) · markDefs (link only) |
| `richTextCode` | language ∈ 15 labels · filename? · code |
| `richTextImage` | asset ref matching the Sanity image-id pattern · alt · caption? |
| `richTextTable` | headerRow · rows of plain-text cells · caption? |
| `richTextCallout` | tone ∈ info/success/warning · title? · body |

Link hrefs: site-relative path, explicit `http(s)://` with a host, or `mailto:`. Size ceilings throughout.

A markDef with a missing or duplicate `_key` is **dropped, never re-keyed** — regenerating it would silently re-point a span's link at a different URL.

## Enforced on both sides

Sanitising on write alone is not enough: Sanity Studio and dataset tooling bypass tRPC entirely. So the schema rejects on the way in, and `sanitizeRichTextContent` runs again at render and never throws.

Link validation was consolidated — three copies of the scheme rule collapsed into one helper. `mailto:` is admitted for prose links only, with a regression test asserting CTA button hrefs still reject it.

## 80 adversarial tests

Not "a sanitiser exists" but "these specific attacks fail": `javascript:` in five casings, tab- and newline-split `java\tscript:`, `data:text/html`, `data:image/svg+xml`, `vbscript:`, `blob:`, `file:`, `//host`, `\\host`, bare `https://`; smuggled `onclick` / `style` / `srcdoc` / `dangerouslySetInnerHTML` on blocks, spans and annotations; unmodelled types including `script`, `iframe`, `__proto__` and `constructor`; ghost marks; duplicate markDef keys re-pointing a link; ten bad image refs including remote URLs, SVG assets and path traversal; prototype pollution; plus DOM-level assertions that no `script`/`iframe`/`object`/`embed`/`img` element is ever produced and that payloads render as visible escaped text.

## Deliberately excluded

- **Syntax highlighting** — a highlighter is a tokeniser that emits markup, which is precisely the machinery this block exists to avoid.
- **Rich text inside callouts and table cells** — recursion widens the attack surface for little editorial gain.
- **SVG images** — an SVG is an active document and the CDN serves it as `image/svg+xml`. Blocked at both the upload MIME allowlist and the asset-id pattern.

## Flagging a pre-existing hole, not touched here

`src/lib/email/portableTextToHTML.tsx` interpolates `value.href` unescaped into an `href` attribute with no scheme check — the same attack class, in the email renderer. It deserves its own change.

## Verified

Light and dark at 393×852 and 1280×900, including a stress story: 200-character single-line code, a 6-column table, a 78-character unbroken link, nested lists. Document `scrollWidth == clientWidth` in every shot — code and tables scroll inside their own boxes, never the page. Shooting the admin editor also revealed the per-row table delete button hiding past the scroll edge on mobile; fixed.

375 files / 4941 tests, typecheck, lint, format, knip clean.

<h3>Greptile Summary</h3>

This PR hardens and extends the homepage rich-text section.
- Adds a strict rich-content vocabulary for prose, code, images, tables, and callouts.
- Sanitizes content at schema validation, editor loading, and public rendering boundaries.
- Adds an organizer-authorized Sanity image upload route and rich-content editor.
- Consolidates safe link handling and adds adversarial schema and DOM tests.

The rich-text hardening appears sound, but the image upload size boundary must be aligned with multipart parsing before merging so valid advertised uploads do not fail.

The new upload route parses the complete request before applying its 8 MB file check, causing images within the advertised range but above the deployment parser limit to fall into the generic error response.

**Files Needing Attention:** src/app/api/admin/rich-text-image/route.ts

| Filename | Overview |
|----------|----------|
| src/lib/homepage/richText.ts | Defines the allowlisted rich-content model, key normalization, limits, and defensive sanitizer used across write and render paths. |
| src/server/schemas/conference.ts | Replaces permissive homepage rich-text validation with a strict discriminated union followed by sanitizer normalization. |
| src/components/homepage/RichTextContent.tsx | Adds render-time sanitization and vetted components for code, images, tables, and callouts. |
| src/components/admin/RichTextContentEditor.tsx | Adds the organizer-facing segmented editor for interleaved prose and rich object blocks. |
| src/app/api/admin/rich-text-image/route.ts | Adds an organizer-scoped image upload endpoint, but its 8 MB application limit exceeds the earlier multipart parsing boundary. |
| src/lib/portabletext/safeHref.ts | Centralizes site-relative, HTTP(S), and rich-text mailto URL validation. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart LR
  A[Organizer editor or Sanity Studio] --> B[Rich-text content]
  B --> C{Write path}
  C -->|Admin editor| D[Zod validation and sanitization]
  C -->|Sanity Studio| E[Stored dataset content]
  D --> E
  F[Organizer image upload] --> G[Authorization and file checks]
  G --> H[Sanity image asset]
  H --> B
  E --> I[Render-time sanitization]
  I --> J[Allowlisted Portable Text components]
  J --> K[Public homepage]
```

<sub>Reviews (1): Last reviewed commit: ["feat(homepage): sanitised rich-content v..."](https://github.com/cloudnativebergen/website/commit/76dd5b200123cbb33f362f27ad7b4dee14ec1d5a) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=49726717)</sub>


Starefossen:
All eight findings were real and are fixed in `c30a3c5f`.

**Upload route (greptile P1 + qodo).** `request.formData()` throws on a body over the deployment's multipart limit *before* the explicit 8 MB check can run, so an organizer whose image was merely too large got told the server broke. Parse failures are now classified: body-size → 413 with an actionable message, anything else rethrown as a genuine 500.

While matching the sibling pattern I found the siblings are themselves subtly broken: `/api/upload/speaker-image` and `/api/admin/gallery/upload` classify with `error.message.includes('body')` — case-sensitively, so they miss the runtime's own `Body exceeded ... limit` and return a 500 on exactly the case that check was written for. Not touched here (out of scope); follow-up PR will lift the classification into one shared helper rather than add a third copy.

**Link message drift (qodo + copilot).** Rich-text links accept `mailto:` but rejection reported `UNSAFE_LINK_MESSAGE`, which describes the stricter CTA rule — organizers were shown a rule other than the one applied. Separate message for the rich-text gate; the CTA constant is untouched, since `mailto:` is deliberately refused there.

**safeHref comment (copilot).** Rather than correct the comment to describe the duplication, the Studio now genuinely imports the shared predicate, so the two can't drift. Verified by building the Studio and grepping the emitted bundle for the message string. That removed a live drift, incidentally: the Studio matched `/^mailto:/i` while the shared module parses the URL and compares `protocol`.

**Validated ≠ sent (copilot).** The real one. `validate` and `toPayload` now receive the *same* sanitized array, computed once, so the invariant holds structurally rather than by two calls happening to agree. And instead of silently dropping a half-finished card behind a "Homepage updated" toast, the save is blocked and the card is named.

**AVIF (copilot ×2).** Both the file input and the Studio `options.accept` now derive from `RICH_TEXT_IMAGE_MIME_TYPES` instead of restating it — a hand-written list is how this drifted in the first place.

**Help text (copilot).** It claimed hostile content is "removed when you save"; the write path rejects it. Now says what actually happens in each case.

**Unbounded marks scan (qodo).** Capped at 50 — deliberately the same limit `richTextSpanSchema.marks` already enforces on the write path, so the render-side cap can only ever truncate data that bypassed the mutation, never anything a legitimate save produced.

**Studio table cells (qodo).** Per-cell length now validated at authoring time against the same `RICH_TEXT_LIMITS.tableCell` the renderer truncates to.

4998 tests green, typecheck clean, knip clean.

PR #709, opened 2026-08-03
fix(branding): derive dark-mode brand tints from the tenant colour

Starefossen:
Dark mode ignored a tenant's brand colour. Every dark-mode brand surface was a hard-coded house blue — `#1E40AF`, `#1E3A8A`, `#2563EB`, `#93C5FD`, `#155E75` — so a conference that stored its own colour got *its* brand in light mode and Cloud Native's in dark mode.

## The transform

Each dark tint is derived from the tenant's own colour by mixing toward black or white **in OKLab** until it hits a fixed target *perceptual lightness*. The five targets are the OKLab lightness values of the five house shades above, so the rule reproduces the hand-tuned palette instead of replacing it.

The house shades aren't one uniform transform in sRGB — blue-700→blue-800 scales by ~0.82, cyan-500→cyan-800 by ~0.55 — but in OKLab they share an invariant: blue-800 sits at L 0.424 and cyan-800 at L 0.450. The hand-tuning was hue compensation for a constant lightness target, and one rule reproduces both.

Not `color-mix`: the required mix percentage is `targetL / L(tenant colour)`, which depends on the tenant's own lightness — something CSS cannot measure. A fixed percentage would reintroduce exactly the hue-dependence being removed.

## Why this is safe for the three live sites

All three existing editions store **no theme**, so any drift would be a live regression. The derivation runs only inside `conferenceThemeCss`, behind an anchored all-or-nothing hex gate, and every changed `.dark` rule keeps a fallback byte-identical to its previous value. Verified with 64 unthemed computed-style probes and 192 themed light-mode probes — zero differences — plus SHA-256-identical Storybook captures.

## What it fixes for tenants

- A **navy** primary (`#0A1F44`) currently renders a near-black, effectively invisible button on `bg-gray-950`. Now a legible mid slate-blue.
- A **bright yellow** primary (`#FACC15`) currently paints white text on full-brightness yellow — about 1.2:1, unreadable. Now a deep olive-gold with a legible white label.
- Storing the **house colours** becomes a genuine no-op: previously it produced a glaring bright-blue→cyan hero where the default has a deep navy→teal one. **This is what unblocks running migration 046**, which pins the house theme onto the three existing editions.

## Review

Independently reviewed with an adversarial pass that re-derived the maths against a separate OKLab/WCAG implementation, using a 140,608-input sweep plus an analytical gamut-boundary scan — bounding the worst case for the entire sRGB cube rather than sampling it. Every load-bearing claim held:

- dark brand text ≥ **7.46:1** on gray-800 (AAA) for *any* tenant primary
- white on the dark brand surface ≥ **7.81:1** (AAA)
- house-colour residuals ΔE 0.009–0.032, lightness exact
- gamut bisection terminates; black/white/near-black collapse to identical neutral greys

Its one finding is fixed here: `parseHex` sliced unvalidated input, so `'#12345'` became `#3b5733` — a silently wrong brand colour. Now rejected.

## Known limits, not introduced by this change

- **Borders** bottom out at 2.50:1 against gray-800, under the 3:1 non-text threshold — the *house* border is already 2.84:1, so this is the existing palette band, not a regression. Raising it would change unthemed pixels.
- **Dark hover has no colour change**, before or after: the house palette paints `bg-brand-cloud-blue` and its hover the same `#1E40AF` in dark mode, so a themed tenant must match or storing a theme stops being a no-op. Giving dark mode a real hover step is a house-palette decision that has to move unthemed pixels first.
- A **pastel** primary yields muted mauve-grey surfaces. Readable — that is what the transform guarantees — but a pastel simply carries little chroma once moved to a dark-surface lightness.
- Three seams still use the light token in dark mode and are left for a follow-up, each needing its own pixel-identity argument: the `Button` outline ring, two `.hover\:*` utilities lacking `.dark` counterparts, and the admin-only slider thumbs.

<h3>Greptile Summary</h3>

Derives perceptually normalized dark-mode brand tints from each tenant’s colours.
- Adds OKLab conversion, gamut mapping, and contrast-focused tests.
- Emits dedicated dark-mode CSS variables from conference themes.
- Rewires several dark brand utilities to derived tokens and expands Storybook coverage.
- Hardens colour parsing against malformed hexadecimal input.

The dark-token derivation is sound, but the remaining variant-only dark utilities must be fixed before merging because existing branded surfaces still render raw light-mode tenant colours.

The new variables work for literal base-class selectors, but concrete badge components use only `dark:bg-brand-cloud-blue`, which continues resolving through the light primary and bypasses the derived dark surface token.

**Files Needing Attention:** src/styles/tailwind.css

| Filename | Overview |
|----------|----------|
| src/lib/branding/color.ts | Adds validated OKLab-based tint derivation with fixed-lightness gamut mapping. |
| src/lib/branding/theme.ts | Generates dedicated dark primary and accent variables for complete tenant themes. |
| src/styles/tailwind.css | Rewires hand-written dark selectors, but variant-only `dark:` utilities still consume light-mode tokens. |
| src/lib/branding/color.test.ts | Thoroughly tests transform accuracy, gamut handling, contrast, formatting, and malformed inputs. |
| src/lib/branding/theme.test.ts | Verifies generated dark declarations but does not cover computed styles for Tailwind dark variants. |
| src/components/BrandThemeShowcase.stories.tsx | Adds representative dark-theme palettes and branded surface examples. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart LR
  A[Tenant primary colour] --> B[conferenceThemeCss]
  B --> C[OKLab lightness transform]
  C --> D[Derived dark CSS variables]
  D --> E[Hand-written .dark base-class rules]
  A --> F[Light --brand-primary token]
  F --> G[Tailwind dark: variants]
  G --> H[Existing variant-only surfaces bypass dark tint]
```

<sub>Reviews (1): Last reviewed commit: ["fix(branding): reject malformed hex inst..."](https://github.com/cloudnativebergen/website/commit/11a881d1a5812e2bade283c61156a0b5d60e1118) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=49746391)</sub>

**Context used:**

### DIFF
--- a/src/components/BrandThemeShowcase.stories.tsx
+++ b/src/components/BrandThemeShowcase.stories.tsx
@@ -49,6 +49,14 @@ function BrandSample({ theme }: { theme?: ConferenceTheme | null }) {
           Bordered callout using the brand colour.
         </p>
       </div>
+
+      <div className="flex h-16 items-center justify-center rounded-xl bg-aqua-gradient text-sm font-semibold text-white">
+        Aqua gradient (accent endpoint)
+      </div>
+
+      <div className="rounded-xl bg-brand-cloud-blue p-4 text-center text-sm font-semibold text-white">
+        Solid brand surface with white label
+      </div>
     </div>
   )
 }
@@ -62,6 +70,38 @@ const PURPLE: ConferenceTheme = {
 const TEAL: ConferenceTheme = {
   primaryColor: '#0D9488',
   accentColor: '#84CC16',
+}
+
+/**
+ * Stress palettes for the DARK tints. The dark brand rules do not use the raw
+ * primary — they derive shades at fixed perceptual lightness (see
+ * `src/lib/branding/color.ts`), and these four sit at the corners of what a
+ * tenant can pick: near-black, near-white, and two colours whose natural
+ * lightness is nowhere near the dark surface band.
+ */
+const NAVY: ConferenceTheme = {
+  primaryColor: '#0A1F44',
+  accentColor: '#334155',
+}
+const YELLOW: ConferenceTheme = {
+  primaryColor: '#FACC15',
+  accentColor: '#FB923C',
+}
+const RED: ConferenceTheme = { primaryColor: '#DC2626', accentColor: '#F97316' }
+const PASTEL: ConferenceTheme = {
+  primaryColor: '#FBCFE8',
+  accentColor: '#A5F3FC',
+}
+
+/**
+ * The house colours stored AS a theme. This must look like the unthemed
+ * Default in both schemes — storing a theme is meant to be a no-op, which is
+ * only true if the dark rules derive their tints instead of using the raw
+ * primary.
+ */
+const HOUSE: ConferenceTheme = {
+  primaryColor: '#1D4ED8',
+  accentColor: '#06B6D4',
 }
 
 const meta = {
@@ -115,5 +155,28 @@ export const ThemedPurpleDark: Story = {
 export const ThemedTeal: Story = { args: { theme: TEAL } }
 export const ThemedTealDark: Story = {
   args: { theme: TEAL },
+  parameters: { theme: 'dark' },
+}
+
+/** House colours stored as a theme — should be indistinguishable from Default. */
+export const ThemedHouseDark: Story = {
+  args: { theme: HOUSE },
+  parameters: { theme: 'dark' },
+}
+
+export const ThemedNavyDark: Story = {
+  args: { theme: NAVY },
+  parameters: { theme: 'dark' },
+}
+export const ThemedYellowDark: Story = {
+  args: { theme: YELLOW },
+  parameters: { theme: 'dark' },
+}
+export const ThemedRedDark: Story = {
+  args: { theme: RED },
+  parameters: { theme: 'dark' },
+}
+export const ThemedPastelDark: Story = {
+  args: { theme: PASTEL },
   parameters: { theme: 'dark' },
 }
--- a/src/lib/branding/color.test.ts
+++ b/src/lib/branding/color.test.ts
@@ -268,3 +268,25 @@ describe('shiftToLightness — malformed input', () => {
     )
   })
 })
+
+/**
+ * Raised in review alongside the malformed-hex guard, and for the same reason:
+ * an out-of-range target does not blow up, it produces a colour that looks
+ * plausible. formatHex clamps, so the caller gets a wrong tint with no signal.
+ */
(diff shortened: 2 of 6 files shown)


Starefossen:
Four findings assessed — one fixed, three declined with evidence.

**Fixed: unvalidated `targetL` (copilot + qodo).** Real, and the same failure mode as the `parseHex` guard already on this branch: an out-of-range or NaN target propagates through the mix and `formatHex` *clamps* it into a plausible-looking colour, so a wrong brand tint ships with no signal. Now throws `RangeError`, with tests for NaN, ±Infinity, out-of-range and both exact boundaries. Every caller passes a `DARK_TINT_LIGHTNESS` constant so it cannot fire in the app — it is there so the next caller gets an error rather than a colour nobody can explain.

**Declining: "Dark variants bypass derived tints" (greptile, P1).** The mechanism is real — a `dark:bg-brand-cloud-blue` utility on an element without the exact base class will not match the handwritten `.dark .bg-brand-cloud-blue` override, and resolves through the light token. But I enumerated every such call site, matching per *utility kind* rather than per line, and the impact is close to nil:

- **8 admin call sites**, including `BadgeManagementClient.tsx:642` — which is the case the finding describes. Admin never receives a tenant theme, so the light token *is* the house colour there. The cited "#FACC15 stays bright yellow" scenario cannot occur on these.
- **9 public call sites**, of which 7 are `/20` opacity washes (`ImageCarousel`, the four PWA banners, `Hero`) and 2 are different tokens entirely (`brand-cloud-blue-light`, `brand-sunbeam-yellow`) that this change never touched.

A 20% opacity wash of the tenant's own colour on a dark surface is arguably the *desired* rendering — the derived tint at 20% would read muddier. Fixing it would require a hand-written `.dark` override per opacity step, brittle in exactly the way this change removes. Recording it as a known limit; worth revisiting if a themed tenant reports it.

**Declining: "expensive exhaustive sweep test" (qodo).** Measured rather than estimated: the entire `color.test.ts` file — 29 tests including the ~29k-call sweep — runs in **136 ms** (288 ms wall, including transform and setup). That is not a material contribution to a 5000-test suite, and the sweep is what proves the contrast guarantees hold across the sRGB cube rather than at sampled points. Shrinking it would trade real coverage for an unmeasurable saving.

For context: this branch also went through an independent adversarial review that re-derived the OKLab maths against a separate implementation and ran a 140,608-input sweep plus an analytical gamut-boundary scan, bounding the worst case for the entire sRGB cube rather than sampling it. Verdict was ship, and the `parseHex` guard already on this branch was its one finding.

PR #710, opened 2026-08-03
fix(og): make social share cards follow the tenant's brand

Starefossen:
Every tenant's Open Graph card rendered in the Cloud Native house blue. A conference that had configured its own brand colour still posted competitor-branded imagery every time anyone shared its site — the most outward-facing un-themed surface in the product.

## The surface

Six routes generate cards; five share `src/lib/og/template.tsx`, and `badge/[badgeId]` carries its own styles and is the only one on the **edge** runtime. That constrained the fix: `conferenceThemeCss` emits CSS custom properties and `color-mix()`, and Satori supports neither, so cards need literal hex. The all-or-nothing pair rule is now extracted once as `resolveBrandPair()` in `src/lib/branding/theme.ts` and shared by `conferenceThemeCss`, `emailBrandColor` and the new `ogBrandColors()` — rather than re-derived per context.

Each caller passes its own house fallback, which is why the badge card keeps its distinct blue→green when unthemed instead of being flattened onto the main gradient.

## Back-compat, measured not argued

The three live editions store no theme, so unthemed output had to be untouched. Both cards were rendered headlessly on `main` and on this branch and hashed:

- unthemed before = unthemed after = **half-theme** after = **malformed-theme** after → `10295e56…`, byte-identical
- speaker card before = after → `7d78e182…`, byte-identical
- **themed** on `main` also hashes `10295e56…` — i.e. a themed conference rendered the house card. That is the bug, demonstrated rather than asserted.

A half-configured theme (primary without accent) and a malformed value both resolve to unthemed and take the verbatim path, so they render the house card rather than anything broken.

## Contrast

These cards bake white text and white decorative fills onto the brand colour, so a light tenant colour would render unreadable. Rather than flipping the foreground — which would mean re-authoring the baked-white icon SVGs and every `rgba(255,…)` overlay — the background is clamped: channels scale proportionally, preserving hue and channel ratios (pure magenta stays pure magenta), until relative luminance ≤ 0.3, i.e. ≥ 3:1 against white. Every text element on these cards is ≥ 24px bold, so the WCAG AA large-text bar applies. Because the sRGB transfer function is convex, clamping both gradient endpoints bounds every interpolated pixel between them — asserted in the tests, not just argued.

Two limitations, stated rather than hidden:
- The house palette is deliberately **grandfathered** above the ceiling (`#06B6D4` is ~2.4:1), because leaving it verbatim is exactly what makes unthemed output byte-identical. A tenant who stores that same cyan does get nudged to `#05A4BF`.
- The 12px sponsor-tier labels on the speaker card clear 3:1 but not 4.5:1. Pre-existing, unchanged, not made worse.

## Known issue, pre-existing and NOT fixed here

Long conference titles **clip at the right edge** — visible identically in the before and after captures, so it is not a regression, but it does mean a long name is cut off on every share. "Cloud Native Days Norway 2026" is in that range. Deliberately left out of this PR so the byte-identity proof above stays clean; follow-up to come.

30 new tests covering the unthemed-verbatim matrix, clamping, hue preservation and the gradient-interior bound. 5014 tests green, typecheck and knip clean.

<h3>Greptile Summary</h3>

Adds tenant-aware Open Graph branding while preserving route-specific house fallbacks.
- Extracts the all-or-nothing brand-pair resolver for site CSS, email, and OG consumers.
- Adds literal OG gradient generation and luminance clamping for tenant colors.
- Applies themed gradients to shared, speaker, and edge-runtime badge cards.
- Adds tests for fallback identity, validation, clamping, hue preservation, and gradient bounds.

The small-text contrast calculation should be corrected before merging because the affected labels render on translucent rather than opaque white backgrounds.

The new text color ceiling guarantees contrast only against pure white, while both current consumers composite their white backgrounds over the tenant gradient and can therefore render valid themed labels below the promised contrast ratio.

**Files Needing Attention:** src/lib/og/brand.ts, src/app/(main)/speaker/[slug]/opengraph-image.tsx, src/app/(public)/badge/[badgeId]/opengraph-image.tsx

| Filename | Overview |
|----------|----------|
| src/lib/branding/theme.ts | Centralizes complete-pair validation and reuses it for CSS and email branding. |
| src/lib/og/brand.ts | Introduces OG color resolution and contrast clamping, but small-text contrast is calculated against the wrong chip background. |
| src/lib/og/brand.test.ts | Thoroughly covers fallback and color arithmetic but does not test contrast after translucent chip-background compositing. |
| src/lib/og/template.tsx | Replaces the shared house gradient with the resolved tenant gradient. |
| src/app/(main)/speaker/[slug]/opengraph-image.tsx | Themes the speaker card and sponsor fallback text; the translucent sponsor chips expose the contrast issue. |
| src/app/(public)/badge/[badgeId]/opengraph-image.tsx | Preserves the badge-specific fallback and adds tenant branding on edge, while its translucent verification pill exposes the contrast issue. |
| src/lib/og/styles.ts | Removes brand-specific colors from the shared layout token collection. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart LR
  Theme[Conference theme] --> Resolve[resolveBrandPair]
  Resolve -->|Valid pair| Clamp[Clamp tenant colors]
  Resolve -->|Missing or malformed| Fallback[Route-specific house pair]
  Clamp --> Brand[ogBrandColors]
  Fallback --> Brand
  Brand --> Shared[Shared OG template]
  Brand --> Speaker[Speaker card]
  Brand --> Badge[Badge card]
```

<sub>Reviews (1): Last reviewed commit: ["fix(og): paint the share cards in the te..."](https://github.com/cloudnativebergen/website/commit/7bd558f966f652e66e4b4b078afe182feb445ed7) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=49765703)</sub>

### DIFF
--- a/src/app/(main)/speaker/[slug]/opengraph-image.tsx
+++ b/src/app/(main)/speaker/[slug]/opengraph-image.tsx
@@ -2,6 +2,7 @@ import React from 'react'
 import { ImageResponse } from 'next/og'
 import { speakerImageUrl } from '@/lib/sanity/client'
 import { STYLES, OG_IMAGE_SIZE } from '@/lib/og/styles'
+import { ogBrandColors } from '@/lib/og/brand'
 import { ogImageMetadata } from '@/lib/og/metadata'
 import { PLATFORM_NAME } from '@/lib/branding/platform'
 import {
@@ -26,6 +27,7 @@ const createSponsorLogo = (
   logoBrightSvg: string | null,
   sponsorName: string,
   isLarge: boolean,
+  textColor: string,
 ) => {
   const commonStyle = {
     padding: isLarge ? '12px 20px' : '8px 16px',
@@ -35,7 +37,7 @@ const createSponsorLogo = (
       : STYLES.borderRadius.tiny,
     fontSize: isLarge ? '16px' : '14px',
     fontWeight: '600',
-    color: STYLES.colors.blue,
+    color: textColor,
     fontFamily: STYLES.fontFamily,
   }
 
@@ -72,8 +74,16 @@ const renderSponsorLogo = (
   logoSvg: string | null,
   logoBrightSvg: string | null,
   sponsorName: string,
-  size: 'small' | 'large' = 'small',
-) => createSponsorLogo(logoSvg, logoBrightSvg, sponsorName, size === 'large')
+  size: 'small' | 'large',
+  textColor: string,
+) =>
+  createSponsorLogo(
+    logoSvg,
+    logoBrightSvg,
+    sponsorName,
+    size === 'large',
+    textColor,
+  )
 
 export function generateImageMetadata() {
   return ogImageMetadata((brand) => `${brand} Speaker Profile`)
@@ -222,6 +232,8 @@ export default async function Image({
   }
 
   const fonts = await loadBrandFonts(domain)
+  // Per-tenant brand gradient; unthemed conferences resolve to the house pair.
+  const brand = ogBrandColors(conference.theme)
   const { speaker, talks, err } = await getPublicSpeaker(
     conference._id,
     decodedSlug,
@@ -236,7 +248,7 @@ export default async function Image({
           display: 'flex',
           alignItems: 'center',
           justifyContent: 'center',
-          background: STYLES.gradient,
+          background: brand.gradient,
           color: 'white',
           fontSize: 48,
           fontWeight: 'bold',
@@ -291,7 +303,7 @@ export default async function Image({
         height: '100%',
         display: 'flex',
         flexDirection: 'row',
-        background: STYLES.gradient,
+        background: brand.gradient,
         color: 'white',
         padding: '40px 40px 80px 40px',
         fontFamily: STYLES.fontFamily,
@@ -407,6 +419,7 @@ export default async function Image({
                     sponsor?.sponsor?.logoBright || null,
                     sponsor?.sponsor?.name || `Sponsor ${index + 1}`,
                     'small',
+                    brand.textOnLight,
                   )}
                 </div>
               ))}
@@ -610,6 +623,7 @@ export default async function Image({
                     sponsor?.sponsor?.logoBright || null,
                     sponsor?.sponsor?.name || `Sponsor ${index + 1}`,
                     'small',
+                    brand.textOnLight,
                   )}
                 </div>
               ))}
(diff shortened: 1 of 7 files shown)


Starefossen:
Both findings are **real, same root cause**, fixed in `aa98d2b7`.

**The arithmetic.** There are exactly three places a brand colour is painted as text — the sponsor-name fallbacks on the speaker card (`:422`, `:626`) and the "OpenBadges 3.0 Verified" pill (`:197`) — and all three sit on `rgba(255,255,255,0.9)`, not white. Satori emits SVG and resvg blends gamma-encoded sRGB directly, so the composite is a plain per-channel lerp; worst case (chip over the dark gradient end) is `0.9 × 255 = 229.5` → relative luminance **0.7835**, not 1.0.

| Text at the ceiling | vs pure white (what the clamp assumed) | vs the real composite |
|---|---|---|
| L = 0.18 (old) | 4.565:1 ✓ | **3.624:1 ✗** |
| L = 0.135 (new) | 5.32:1 ✓ | **4.506:1 ✓** |

So the guarantee genuinely did not hold. `OG_MAX_TEXT_LUMINANCE` is now 0.135, derived as `(0.7835 + 0.05) / 4.5 − 0.05`.

I fixed the **ceiling**, not the composite: making the chips opaque white would change unthemed bytes and break the back-compat proof.

**The size assumption was right, though.** Those chips are 14–16px at weight 600, which is *not* WCAG "large" (needs ≥24px, or ≥18.66px bold), so 4.5:1 was correctly the bar. The defect was purely the background it was measured against.

The derivation is now executable code (`compositeWhiteOver`, `contrastRatio`, `OG_CHIP_BACKGROUND_ALPHA`) that the tests recompute, rather than a number justified in prose. 30 → 43 tests, including an explicit regression test proving the old 0.18 gave 3.62:1, and a sweep over every gradient pixel behind a chip. Real brand colours are below both ceilings and still returned verbatim; only pastels move.

**Byte-identity re-verified** three ways within one fixture — pre-theming `main`, this branch before the fix, and after it — identical for both the root and speaker cards. The unthemed path returns the fallback and never consults the ceiling, so this is structural rather than incidental.

---

**A third issue, surfaced but deliberately NOT fixed — flagging for a decision.**

`OG_MAX_BACKGROUND_LUMINANCE`'s doc claimed "every text element on these cards is ≥ 24px bold". That is **false**: there is white-on-gradient text at 12px/600 (sponsor-section labels), 16px/600 (badge domain) and 18px/700 (date and location). Those need 4.5:1, which would require a background ceiling of ≤ 0.1833 — the current 0.3 only delivers 3:1.

I left 0.3 alone because it is inherited from the house design rather than introduced here, and because the house cyan `#06B6D4` (L = 0.3825) gives those elements **2.43:1 today — worse than any themed tenant can be**. Tightening it would darken every themed gradient, which is a design decision, not a correctness fix. I corrected the doc to state the truth and pinned it in a test so the overclaim cannot regrow.

PR #711, opened 2026-08-03
feat(homepage): the section variant registry

Starefossen:
Foundation for per-section variants — the mechanism all 13 section types will follow, so the front page can offer a choice of how each section LOOKS, not just what it says.

Purely additive. Nothing consumes the registry yet, so this changes no rendering and no stored document.

## What it adds

`src/lib/homepage/variants.ts` — a CLOSED `SECTION_VARIANTS` registry where the first entry per type is the default, plus `defaultVariant()`, `resolveVariant()` and a type guard. Unknown variants fall back to the default with a warn-once, a softer sibling of the renderer's existing unknown-`_type` skip: content still renders rather than vanishing.

| Type | Variants (first = default = today) |
|---|---|
| Hero | `classic`, `minimal`, `emblem` |
| Save the date | `card`, `strip` |
| Featured speakers | `shelf`, `grid` |
| Program highlights | `full`, `talks` |
| Organizers | `cards`, `compact` |
| Sponsors | `tiers`, `logo-wall` |
| Gallery | `carousel`, `mosaic` |
| Metrics | `row`, `band` |
| CTA banner | `plain`, `panel` |
| Rich text | `article`, `boxed` |
| FAQ | `accordion`, `list` |
| Countdown | `units`, `strip` |
| Venue | `card`, `split` |

Four types get exactly two variants rather than three — metrics, countdown, venue and program highlights genuinely have little meaningful variation, and padding the matrix would produce options nobody wants.

## Back-compat is structural, not tested-in

The default variant is **never persisted**. The three live editions' documents are untouched, and an absent variant resolves to the first entry — which is today's rendering. There is no migration and nothing to roll back.

## Zero runtime dependencies

`variants.ts` has no imports at all, and `sections.ts` imports only its *type*. That is enforced by a source-text test, not convention — a `@dnd-kit` import in the neighbouring `editor.ts` recently broke the production build with `createContext is not a function` during page-data collection, which typecheck and 4900 unit tests could not see. Both edges of this module boundary are now erased at compile time.

205 tests green in `src/lib/homepage` (+12), typecheck and knip clean.

<h3>Greptile Summary</h3>

Adds a closed, dependency-free homepage section-variant registry.
- Defines variants, labels, descriptions, defaults, validation, and fallback resolution for all 13 section types.
- Extends section interfaces with optional, type-specific variant fields.
- Re-exports the registry through the homepage barrel.
- Adds contract, fallback, typing, and server-safety tests.

The PR appears safe to merge with no actionable defects identified in the additive registry foundation.

The registry covers every existing homepage section type, preserves absent-value defaults, keeps section-to-variant typing aligned, and introduces no runtime dependency or active rendering and persistence behavior.

| Filename | Overview |
|----------|----------|
| src/lib/homepage/variants.ts | Adds the complete typed registry and dependency-free resolution helpers; no actionable defect was identified. |
| src/lib/homepage/sections.ts | Adds optional section-specific variant typing consistently across all 13 section interfaces. |
| src/lib/homepage/variants.test.ts | Covers registry completeness, metadata alignment, defaults, validation, fallback warnings, typing, and runtime-import safety. |
| src/lib/homepage/index.ts | Exposes the new dependency-free registry through the existing homepage barrel. |

</details>

<sub>Reviews (1): Last reviewed commit: ["feat(homepage): add the closed per-secti..."](https://github.com/cloudnativebergen/website/commit/c29c051e23b128f654b826203641fcb1a1704bc2) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=49766580)</sub>

### DIFF
--- a/__tests__/helpers/moduleImports.test.ts
+++ b/__tests__/helpers/moduleImports.test.ts
@@ -1,0 +1,175 @@
+import { describe, expect, it } from 'vitest'
+import { findRuntimeModuleImports } from './moduleImports'
+
+/**
+ * The guard this file covers replaced a source-text regex that was copied into
+ * two test files and matched only `import … from '…'` with SINGLE quotes.
+ *
+ * `OLD_GUARD` is that exact regex, kept here as an executable record of what it
+ * missed. Every case below asserts BOTH that the parser sees the import and
+ * that the regex did not — so the fix is demonstrated, not just claimed.
+ */
+const OLD_GUARD = (source: string): string[] =>
+  Array.from(
+    source.matchAll(/^import\s+(?!type\b)[^;]*?from\s+'([^']+)'/gm),
+    (match) => match[1],
+  )
+
+const specifiers = (source: string) =>
+  findRuntimeModuleImports(source).map((i) => i.specifier)
+
+describe('findRuntimeModuleImports — forms the old regex was blind to', () => {
+  const blindSpots: Array<{
+    name: string
+    source: string
+    specifier: string
+    kind: string
+  }> = [
+    {
+      name: 'side-effect import',
+      source: `import '@dnd-kit/core'\n`,
+      specifier: '@dnd-kit/core',
+      kind: 'side-effect import',
+    },
+    {
+      name: 'named re-export',
+      source: `export { arrayMove } from '@dnd-kit/sortable'\n`,
+      specifier: '@dnd-kit/sortable',
+      kind: 're-export',
+    },
+    {
+      name: 'star re-export',
+      source: `export * from '@dnd-kit/core'\n`,
+      specifier: '@dnd-kit/core',
+      kind: 're-export',
+    },
+    {
+      name: 'dynamic import',
+      source: `export async function load() {\n  return import('@dnd-kit/core')\n}\n`,
+      specifier: '@dnd-kit/core',
+      kind: 'dynamic import',
+    },
+    {
+      name: 'double-quoted import',
+      source: `import { DndContext } from "@dnd-kit/core"\n`,
+      specifier: '@dnd-kit/core',
+      kind: 'import',
+    },
+    {
+      name: 'double-quoted side-effect import',
+      source: `import "@dnd-kit/core"\n`,
+      specifier: '@dnd-kit/core',
+      kind: 'side-effect import',
+    },
+    {
+      name: 'require call',
+      source: `const dnd = require('@dnd-kit/core')\n`,
+      specifier: '@dnd-kit/core',
+      kind: 'require',
+    },
+    {
+      name: 'import-equals',
+      source: `import dnd = require('@dnd-kit/core')\n`,
+      specifier: '@dnd-kit/core',
+      kind: 'import =',
+    },
+    {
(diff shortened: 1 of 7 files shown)


Starefossen:
Four findings assessed in `a3077fcd` — three real and fixed, one refuted with build evidence.

**1. Incomplete import guard (copilot + qodo). Real, and worse than reported.**

This one mattered: the guard exists because a `@dnd-kit` import in the sibling `editor.ts` broke the **production build** with `createContext is not a function` during page-data collection — a failure typecheck and ~5000 unit tests could not see. A guard with holes is worse than no guard, because it reads as protection that isn't there.

The regex had three classes of defect, not one:
- the reported missing forms — side-effect `import 'x'`, `export … from`, `export *`, `import()`, double quotes — plus `require()`
- **false positives**: `^import` matches at column 0 *inside a block comment*, so a commented-out import was reported as real
- **cross-statement welding**: `[^;]*?` spans newlines, so one line's `import` welded to the next line's `from './barrel'` and reported a specifier no single statement contains — while still missing all four real imports in the same snippet

Replaced with a **TypeScript AST walk** (`__tests__/helpers/moduleImports.ts`). `typescript` is already a dependency and is the same parser that compiles these files, so the guard's answer is the compiler's answer. An ESLint rule was rejected because it cannot see dynamic `import()` and is currently broken in worktrees; asserting on built output works but costs a multi-minute build, which is the slow loop this guard exists to shortcut.

Proven in situ — each form appended to the real `variants.ts` and the guard re-run. Old regex flagged **none** of the six; the new guard fails on all six.

Both copies now call one helper, so the next copy cannot drift. The old regex survives only inside the helper's own test, as the executable record of what it missed.

**2. Barrel export evaluates variants (qodo). Refuted, with build evidence.**

`variants.ts` has zero runtime imports (now mechanically proven), so the re-export adds three object literals. The barrel already re-exports `./sections` and `./richText`, both of which *do* have runtime imports — variants is its cheapest member. Empirically: `logo-wall`, a string existing only in the registry, appears in **zero bytes** of `.next/server` and `.next/static` after a real production build. Fully tree-shaken.

**3. Unbounded warn cache (qodo). Real, capped.**

`VARIANT_WARN_LIMITS` in the `RICH_TEXT_LIMITS` house style. Bounded in **both** directions — memory by the pair cap, and log volume by emitting one "suppressing further" notice then going silent, since unbounded log spam from a hostile dataset is the same denial-of-service by another route. Long values are clipped so a single entry is bounded too.

**4. Narrow input type (qodo). Real, widened to `unknown`.**

Following the `isBlank()` precedent from migration 046, a wrong-typed value is not folded into the benign case. `undefined` → default, silently: that is the only shape an absent field takes and the only shape the write path can produce. Anything else — `null`, numbers, objects — renders the same safe default but is **reported by type** (`[number]`, `[object]`) rather than quoted as a variant name, so numeric junk isn't disguised as a typo.

Production build exit 0. 225 tests green in the touched areas; 5194 across the suite.

Unrelated, but flagging: the PWA-logomark rasterization test failed once and passed on re-run with no code change — an order-dependent flake, untouched by this work and worth a separate look.

PR #717, opened 2026-08-03
feat(homepage): a content-status map the preview can trust

Starefossen:
Foundation for the front-page composer's live preview (batch E1). Two new files, nothing else touched.

Answers, for each of the 13 section types against a given conference: does this band have what it needs, will it SELF-HIDE on the live site, how much content does it have, and where is that content edited.

## Why it has to be exact

If the status says a band renders and the real renderer hides it, the preview lies to the organizer — which is worse than having no preview. So the guards were derived from the renderer's code, and equivalence is **pinned by rendering**: the test drives 32 cases across all 13 types through the *real* `HomepageSectionRenderer` with the *real* section components (only `next/link` is stubbed, since it needs a router — stubbing a section would make the assertion vacuous) and asserts `(rendered HTML is empty) === willHide`, in both directions. The case table is a `Record<HomepageSectionType, Case[]>`, so a 14th section type fails to compile rather than going silently uncovered.

## Two live-site bugs found while mapping the guards

Not caused by this work, not fixed here — reported so they are not lost:

1. **Sponsors with no sponsors renders an EMPTY BAND, not nothing.** `Sponsors.tsx` has no early return, so a conference without sponsors ships ~10rem of blank whitespace on its front page. Same shape post-event, where the pitch is suppressed regardless of `showCta`. Reported here as `degraded` rather than hidden, because that is the truth.
2. **A CTA banner with only a label or only a URL renders a banner with no way out** — half a button produces a dead call-to-action.

## Other guards worth knowing

- The **countdown has two guards in two files**, and the second runs only after hydration: a passed target server-renders placeholder dashes, then hides on the client. What a visitor ends up seeing is nothing, so `willHide` is true — a server-render-only parity check would have disagreed.
- **Rich text hides on non-empty stored content**: the guard is on the *sanitized* array, so blocks the allowlist drops still hide the band. A naive `content.length > 0` predicate lies here.
- **Sponsors without a tier never render**, and all `special` tiers collapse under one heading — so counts must be computed the way the band groups, not the way the array reads.

## One deliberate deviation

The module takes exactly one runtime import: the sibling `./richText`. The rich-text guard is "does the sanitized content render anything", which is ~200 lines of allowlist logic — copying it here would create precisely the drift this module exists to prevent. `richText.ts` is itself package-free, and `editor.ts` takes the same import for the same reason. The purity test is stricter than its siblings' as compensation: it asserts the import list equals exactly `['./richText']`, so nothing else can slip in.

278 tests green, typecheck, eslint and knip clean.

<h3>Greptile Summary</h3>

Adds a centralized homepage content-status map and renderer-backed parity tests.
- Reports readiness, self-hiding behavior, counts, summaries, and management destinations for all 13 homepage section types.
- Covers time-dependent, sanitized rich-text, sponsor-tier, FAQ-source, and lifecycle-related behavior with real section rendering.

The sponsor count discrepancy should be fixed before merging because the new status map can overstate live content for partially tiered sponsor collections.

The visibility guards and management routes align with their live counterparts, but sponsor status counts every stored sponsor while the live band renders only tiered entries.

**Files Needing Attention:** src/lib/homepage/contentStatus.ts

| Filename | Overview |
|----------|----------|
| src/lib/homepage/contentStatus.ts | Adds statuses and content-source mappings for every homepage section type, but mixed tiered and untiered sponsors produce an inflated ready count. |
| src/lib/homepage/contentStatus.test.tsx | Adds comprehensive renderer-parity and metadata tests, though the sponsor cases omit the mixed tiered/untiered state that exposes the count discrepancy. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart LR
  S[Homepage section] --> C[sectionContentStatus]
  D[Conference content] --> C
  C --> H[Hide or render status]
  C --> N[Count and summary]
  C --> M[Management destination]
  S --> R[HomepageSectionRenderer]
  D --> R
  R --> P[Parity tests]
  H --> P
```

<sub>Reviews (1): Last reviewed commit: ["feat(homepage): content status for every..."](https://github.com/cloudnativebergen/website/commit/a27d08b44a326279a9585639b78e18aff22d5701) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=49813507)</sub>

### DIFF
--- a/src/lib/homepage/contentStatus.test.tsx
+++ b/src/lib/homepage/contentStatus.test.tsx
@@ -1,0 +1,812 @@
+/**
+ * @vitest-environment jsdom
+ */
+import { afterEach, describe, expect, it, vi } from 'vitest'
+import { cleanup, render } from '@testing-library/react'
+import { readFile } from 'node:fs/promises'
+import { join } from 'node:path'
+import { findRuntimeModuleImports } from '../../../__tests__/helpers/moduleImports'
+
+// The ONLY stub in this file, and it is not a section component: `next/link`
+// needs a router in jsdom. Everything the renderer maps to is the REAL
+// component — stubbing one would make the parity assertion below meaningless,
+// because a stub cannot self-hide the way the thing it stands in for does.
+vi.mock('next/link', () => ({
+  __esModule: true,
+  default: ({
+    href,
+    children,
+    ...rest
+  }: {
+    href: string
+    children: React.ReactNode
+  }) => (
+    <a href={href} {...rest}>
+      {children}
+    </a>
+  ),
+}))
+
+import { HomepageSectionRenderer } from '@/components/homepage/SectionRenderer'
+import {
+  CONTENT_SOURCES,
+  sectionContentStatus,
+  type SectionContentSourceId,
+} from './contentStatus'
+import {
+  HOMEPAGE_SECTION_TYPES,
+  type HomepageSection,
+  type HomepageSectionType,
+} from './sections'
+import type { Conference } from '@/lib/conference/types'
+
+afterEach(cleanup)
+
+/**
+ * A conference with NOTHING on it — the day-one state, which is the state that
+ * makes every guard in `contentStatus.ts` fire. Cases below add exactly the one
+ * field they are about, so a passing render can never be explained by a
+ * neighbouring field left set in the fixture.
+ */
+function bareConference(overrides: Partial<Conference> = {}): Conference {
+  return {
+    _id: 'conf-1',
+    title: 'Test Conf',
+    city: '',
+    registrationEnabled: false,
+    organizers: [],
+    ...overrides,
+  } as unknown as Conference
+}
+
+/** Dates chosen so the REAL clock puts them unambiguously in the past/future. */
+const PAST = '2000-01-01'
+const FUTURE = '2099-01-01'
+
+function scheduleWith(status: 'confirmed' | 'submitted') {
+  return [
+    {
+      _id: 's1',
+      date: PAST,
+      tracks: [
+        {
+          trackTitle: 'Track 1',
+          trackDescription: '',
+          talks: [
+            {
+              startTime: '09:00',
+              endTime: '09:45',
+              talk: {
+                _id: 't1',
+                status,
+                title: 'A talk',
+                format: 'presentation_25',
+                speakers: [],
+                topics: [],
+              },
+            },
+          ],
+        },
+      ],
+    },
+  ] as never
+}
+
+interface Case {
+  /** Names the case in the test output. */
+  name: string
+  section: HomepageSection
+  conference: Conference
+  /** The expectation this file is about. Verified against a REAL render. */
+  willHide: boolean
+}
+
+/**
+ * The case table, keyed by section type.
+ *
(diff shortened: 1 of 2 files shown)


Starefossen:
Both findings real, fixed in `eb07a818`.

**Sponsor counts over-report (greptile P1 + qodo) — real, and broader than the flagged line.**

`groupSponsorsByTier` (`src/lib/sponsor/utils.ts:54-74`) skips any sponsor with a falsy tier, and `Sponsors.tsx` renders logos only from that grouping — so a conference with 10 sponsors, 3 untiered, was told "10 sponsors" over a band drawing 7.

Checking the whole counting path rather than the one line found the actual defect was wider: with *some* tiered and *some* untiered the old code fell through to `ready` with the raw total, so the mixed case had no handling at all. The reviewers' line was one symptom of it. The tier count, by contrast, was already correct — refuted as a bug, but folded into the same walk so the two numbers cannot diverge again. One extra fidelity fix: upstream uses `tier.title` as an object key, so the tier set now coerces the same way.

A partially-tiered band now reports `degraded` with a reason naming the excluded count, rather than quietly inflating.

**Return type lied for unknown `_type` (copilot) — real.** `String(unknown?._type) as HomepageSectionType` produced a value that silently misses any `Record<HomepageSectionType, …>` lookup — an `undefined` label in the composer rail. Now a discriminated union on `known`, with the unknown case carrying `type: string`. No new warn was added: the renderer already warns once per process for that `_type` and the composer surfaces it as a row, so a third pattern would just be noise.

**How it is verified.** The new test does not restate the grouping rule — it renders through the real `HomepageSectionRenderer`, counts the logo links and tier headings out of the DOM, and asserts the reported count and summary match what was actually drawn. Reintroducing the bug (`shown = sponsors.length`) fails exactly those two tests with `expected 5 to be 3`, and the finding-2 tests do not compile against the old signature.

283 tests green; the import-list assertion still holds at exactly `['./richText']`.

Starefossen:
Remaining findings addressed.

**Double schedule traversal (qodo) — real, fixed.** The programme branch computed `tallyProgramme` and then called `hasProgramme`, which computed it again — two full walks of every track of every schedule to answer "> 0", a question the first walk had already answered. `hasProgramme` now takes the session count.

**Earlier findings on this PR, for the record**, since the bot threads stay open once posted:
- *Sponsor counts over-report* — fixed, and the defect was wider than the flagged line: the mixed tiered/untiered case had no handling at all.
- *Metrics count* — fixed. `MetricsBlock` draws the first six entries and filters nothing, so the count over-promised twice: entries past the sixth are stored and never drawn, and an entry with neither label nor value spends one of the six cells drawing nothing.
- *Programme talks counted twice* — the count itself was right (it matched the band's own "N+ Sessions" tile); the **label** was the lie. One talk booked into two slots renders two cards, and calling that "2 confirmed talks" invented a talk the conference does not have. `count` is now sessions, with the distinct talks named in the summary when they differ.
- *Return type lied for unknown `_type`* — fixed with a discriminated union on `known`.

Every count assertion is derived from the rendered DOM — metric cells that came out with text, the Sessions number parsed back out of the band's own stat tile — rather than restating the rule in the test, which is how the original sponsor bug slipped through. 290 tests green; the module's import list is still exactly `['./richText']`.

PR #718, opened 2026-08-03
feat(homepage): sample content so a new conference can see its front page

Starefossen:
Foundation for the front-page composer's live preview (batch E2).

## The problem

A conference configuring its front page early has no speakers, no sponsors, no gallery and no schedule. Several bands self-hide when empty, so a live preview would render nearly blank — exactly when the organizer most needs to see what they are building.

## The design constraint that shaped it

Placeholders make the preview useful; the risk is an organizer believing they have twelve speakers. So: **no invented people.** Speakers and organizers get initials-avatars, not plausible names with stock faces. Sponsors get generated SVG wordmarks reading `SAMPLE A…F`. Every metric label carries the word "Sample" so even a stray screenshot reads as fake.

Only EMPTY collections are filled — a conference with 2 real speakers shows 2, never 2 plus 10 invented ones. The module returns the identical object reference when nothing was empty, so callers can skip cheaply.

Two seams let the UI mark them, neither string-based: `placeholderTypes` (band level, for the "Sample content" chip) and `isPlaceholder(entity)` — a plain property rather than a symbol, deliberately, so it survives `structuredClone`/`postMessage`.

Placeholders honour the tenant theme. Inline SVG uses the `--brand-*` custom properties and re-skins live; data-URI SVG inside `` cannot read the host document's custom properties, so those bake the conference's own hex using the same all-or-nothing pair rule as `conferenceThemeCss` — pinned by a test against the real constants.

## Purity: placeholder bytes must never reach the public page

- Zero runtime imports, asserted with the shared AST guard.
- A test walks the **actual import graph** from every non-admin file under `src/app` (~180 entries, >300 reachable modules), resolving `@/` and relative specifiers, and asserts this module is never reached — printing the import chain on failure. Two anti-vacuity guards: a floor on the reachable count, and a positive control proving the walker *does* find `sections.ts` from the public homepage.
- Production-build grep: attempted and reported honestly — the build compiled, then failed page-data collection on an unrelated missing `RESEND_API_KEY` in the worktree; `grep -r konf-placeholder .next` over 164 MB of output returned zero hits. Caveat: nothing imports the module yet, so that grep is nearly free today. The AST graph test is the guard that keeps mattering.

## Note for the composer batch

Gallery tiles carry a well-formed but nonexistent Sanity asset ref, because `@sanity/image-url` **throws** on a malformed ref and `ImageCarousel` calls it unconditionally — a malformed one would take the gallery band down inside the preview. The honest artwork is the gradient data URI; rendering through the CDN builder yields a 404 and the carousel's own error state.

257 tests green (32 new), typecheck, eslint, prettier and knip clean.

<h3>Greptile Summary</h3>

Adds a preview-only placeholder data generator for otherwise empty homepage sections.
- Generates marked sample speakers, organizers, sponsors, gallery entries, schedule content, metrics, FAQs, venue details, and dates.
- Adds deterministic theming, SVG generation, placeholder detection, and public-import-graph safeguards.
- Adds extensive tests covering content preservation, rendering contracts, determinism, theme behavior, and module isolation.

The gallery placeholder path needs to be fixed before merging because the intended preview renders error states instead of the generated sample artwork.

Gallery placeholders deliberately reference nonexistent Sanity assets while the existing renderer ignores their gradient `imageUrl`, so the newly generated gallery cannot fulfill its preview contract.

**Files Needing Attention:** src/lib/homepage/placeholders.ts

| Filename | Overview |
|----------|----------|
| src/lib/homepage/placeholders.ts | Adds comprehensive placeholder generation, but gallery placeholders store their visible artwork in a field the existing renderer never consumes. |
| src/lib/homepage/placeholders.test.ts | Adds broad behavioral and import-graph coverage, although the gallery test only verifies that the fake Sanity reference does not throw and misses the resulting render-time 404. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart LR
  A[Empty conference collections] --> B[withPlaceholders]
  B --> C[Generated placeholder conference]
  C --> D[HomepageSectionRenderer]
  D --> E[ImageGallery]
  E --> F[ImageCarousel]
  F --> G[Nonexistent Sanity asset URL]
  G --> H[Failed to load image state]
  C -. unused imageUrl gradient .-> F
```

<sub>Reviews (1): Last reviewed commit: ["feat(homepage): sample content for the c..."](https://github.com/cloudnativebergen/website/commit/38613dec58abeaec6e79758e28879d20f1d9731b) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=49813520)</sub>


Starefossen:
Four findings, all real, fixed in `38bd48e0` — except one that cannot be fixed inside this module, explained below.

**Gallery gradients never render (greptile P1) — real, and the fix needs a change elsewhere.**

Traced it properly rather than patching around it: `SectionRenderer` → `ImageGallery` → `ImageCarousel`, which builds both `src` and `srcSet` from `sanityImage(currentImage.image)` and never reads `imageUrl` (`ImageCarousel.tsx:122-141`; same pattern in `GalleryModal.tsx:180,326` and `SimpleImageCarousel.tsx:84`). The placeholder's deliberately-dead ref therefore 404s and the carousel shows its error panel.

Every escape hatch inside `@sanity/image-url` was checked before concluding: `parseSource` accepts ref / asset / asset-stub-with-url / string / URL-string, but **every** path ends in `specToImageUrl`, which composes a CDN URL — there is no source shape that returns a caller-supplied `data:` URI. The single `data:` return in the library is a hardcoded 1×1 transparent PNG for in-progress uploads. And simply dropping the `image` field is worse: the carousel guards on `currentImage?.image`, so an absent field renders no `` at all and leaves a spinner that never stops.

So this module now states an explicit two-sided **gallery contract** — `imageUrl` is the artwork (mirroring the real GROQ projection `"imageUrl": image.asset->url`, so it exists on real images too), `image` is a decoy that exists only to stop the builder throwing, and consumers must prefer `imageUrl` when it is a `data:` URI. Two tests pin both halves, including one asserting the CDN builder *cannot* produce the artwork, so the constraint cannot quietly rot.

The remaining three lines belong to the gallery components, which another branch owns: mirror the existing `speakerImageUrl` pattern (`src/lib/sanity/client.ts:45-64`) so those three call sites use `image.imageUrl` when it starts with `data:` and fall through to `sanityImage(...)` otherwise. Tracked and dispatched separately — deliberately not smuggled in here, and it introduces no import from this module, so the import-graph purity test stays green.

**Eager construction (qodo) — real.** `sampleSpeakers` ran unconditionally at the top of `withPlaceholders`. Now memoised and called only from the branches that need it, so a fully-populated conference builds no speakers, no avatars and no SVG.

**Unmarked schedule wrappers (qodo) — real, fixed by marking rather than by weakening the contract.** Tracks and talk slots were unmarked; so were the sponsor and tier records nested inside `ConferenceSponsor`, which is the same class of gap and arguably worse, since a sponsor wall renders the nested record rather than the wrapper. The contract now says "every generated **object**", with an explicit carve-out for plain field values, and a test deep-walks a fully-filled conference and fails with the path of any unmarked object — so code and contract cannot drift apart by spot-check again.

**`Array.shift()` BFS (qodo, informational) — real.** Now a read cursor. The graph walk still asserts >300 reachable modules.

Purity untouched: zero runtime imports, and the module remains unreachable from every non-admin route. 261 tests green.

PR #724, opened 2026-08-03
perf: stop preloading three fonts nothing renders

Starefossen:
Found during a typography audit; worth shipping on its own.

The root layout preloads six Google font families. Three of them — Bricolage Grotesque, IBM Plex Sans and IBM Plex Mono — have **zero usages in shipped page rendering**. Their only references are Storybook stories under `src/docs/` (imported by nothing under `src/app` or `src/components`) and a font-name dropdown in the admin meme generator.

Three `preload: false` lines. Nothing deleted, so the single admin call site keeps working.

## Measured, not estimated

Full production build before and after, mapping every emitted `.woff2` back to its `@font-face` family through the built CSS:

| | Preloaded files | Preloaded bytes |
|---|---|---|
| Before | 9 | 233,068 B |
| After | 3 | 111,232 B |
| **Saved** | **6** | **121,836 B (~119 KB, 52.3%)** |

Total emitted font bytes are unchanged (45 files, 690,364 B) — intended, since nothing was removed, only de-prioritised.

Note the audit that surfaced this estimated ~300 KB. The real figure is ~119 KB — about 2.5× smaller. Still half the preload budget on every page load of every tenant site.

## One consequence, flagged rather than fixed

`MemeGenerator.tsx:300` sets `ctx.font` on a canvas, and canvas text does **not** trigger a lazy webfont download (there is no `document.fonts.load()` anywhere in the repo). So on `/admin/marketing`, choosing Bricolage or IBM Plex for a meme may now fall back to the system sans unless the font happens to be cached. `preload: true` was masking that by accident.

Admin-only and cosmetic, and the picker is already best-effort — it offers "Atkinson Hyperlegible", which was never loaded via `next/font` at all. The proper fix is `await document.fonts.load(...)` before drawing; out of scope here.

Typecheck, eslint and production build clean. Two `__tests__/lib/pwa/` icon-rasterization tests time out under full-suite parallel load; verified unrelated — they pass in isolation both with and without this change.

<h3>Greptile Summary</h3>

This PR reduces global font preloading while retaining all font declarations.
- Disables eager preloading for IBM Plex Sans, IBM Plex Mono, and Bricolage Grotesque.
- Keeps their CSS variables and font-face resources available for lazy use.

The font preload optimization needs a corresponding explicit font-load step in the meme generator before it is safe to merge.

The three fonts remain selectable by a shipped canvas feature, but disabling their only eager load while drawing synchronously allows the selected family to be replaced by a fallback in generated output.

**Files Needing Attention:** src/app/layout.tsx and src/components/admin/MemeGenerator.tsx

| Filename | Overview |
|----------|----------|
| src/app/layout.tsx | Disables three unused page-rendering preloads, but this makes the existing synchronous meme-generator canvas path render selected fonts unreliably. |

</details>

<sub>Reviews (1): Last reviewed commit: ["perf(fonts): stop preloading the three u..."](https://github.com/cloudnativebergen/website/commit/1783d9c53f54a90b0a5021e4effb855fda2d0e47) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=49828972)</sub>


Starefossen:
The canvas-font consequence is fixed in `b162d63c` rather than shipped as a known regression — three reviewers were right to push back on deferring it. The `preload: false` change is untouched.

**The fix.** `ctx.font` is now built from a shared helper, and an effect requests the faces the current text needs and redraws once they land. The synchronous draw still runs first, so the canvas paints in the fallback immediately and upgrades a moment later — **nothing blocks on the font load**. That is why no timeout is needed: a face that 404s, does not exist, or never settles leaves the preview exactly as it renders today. Requests are keyed on family/weight/size/text, so colour and position edits do not refire loads.

**Verified against a real browser, not asserted.** Playwright against `next dev` (deliberately not Storybook, which loads fonts from the Google CDN and would prove nothing about `next/font`):

- `next/font` **does** register the literal family names — the hash only lands in the class name, so the picker's family strings are correct.
- On a fresh page load every one of those faces is `status: "unloaded"`, confirming the P1 was real rather than theoretical.
- After the load call: the matched face flips to `loaded`, a new woff2 appears on the wire, and canvas `measureText` moves **1437.89 → 1358.28 px**. The metrics change is the part that proves it is not a silent no-op — the usual failure mode here is a wrong shorthand that resolves without loading anything.

**One trap worth recording:** `document.fonts.check()` is not a valid gate — it returns `true` for a family with no `@font-face` at all (verified with a nonsense family name). The code does not use it.

**Atkinson Hyperlegible** has no `@font-face` anywhere in the app — only a token resolving to `system-ui` and a Storybook-only CDN link. So it was always best-effort and stays so: the load resolves with an empty match list, no throw, no hang, and the draw proceeds in the fallback.

**No other client-side canvas path has the same assumption.** The signature pad and upload zone draw strokes and images only; the sharing images use html2canvas over real DOM, which triggers the lazy fetch normally and uses still-preloaded families; badge PNG and OG images are server-side with explicit font files. `MemeGenerator` was the only client-side `ctx.font` in the repo.

The component also had no story before this; it has two now, with a negative control proving the font assertion is not vacuous.

PR #725, opened 2026-08-03
fix(gallery): render images whose artwork is a data: URI

Starefossen:
The gallery components build image `src`/`srcSet` exclusively from the Sanity CDN URL builder and never consult `imageUrl`. That is fine for uploaded photos, but any image whose artwork is a `data:` URI cannot render at all — the builder composes a CDN URL from the asset ref, which 404s, and the carousel falls into its "Failed to load image" state.

This blocks the front-page composer's live preview, where placeholder gallery tiles are generated gradient artwork. It is also a latent gap in its own right: `imageUrl` is a real projected field (`"imageUrl": image.asset->url`), present on real images, that nothing in the gallery path has ever used.

## Why it could not be fixed inside the image builder

Every `parseSource` path in `@sanity/image-url` ends in `specToImageUrl`, which composes a CDN URL — there is **no** source shape that returns a caller-supplied `data:` URI. The library's single `data:` return is a hardcoded 1×1 transparent PNG for in-progress uploads. And simply omitting the `image` field is worse: the carousel guards on `currentImage?.image`, so an absent field renders no `` at all and leaves a spinner that never resolves.

## The fix

`galleryImageSrc()` beside the existing `speakerImageUrl` in `src/lib/sanity/client.ts` — which already solves exactly this problem for speaker photos, and is the precedent this mirrors. Returns `imageUrl` verbatim when it is a `data:` URI, otherwise applies the existing builder chain unchanged. Wired at all four call sites: `ImageCarousel` (src + srcSet), `GalleryModal` (main image + thumbnail), `SimpleImageCarousel`.

`srcSet` is omitted for `data:` URIs: there are no responsive variants of inline bytes, so a srcSet could only repeat the same payload at 1x and 2x, doubling the inline bytes in the DOM for nothing.

## Evidence real images are unchanged

There were **no existing tests for any gallery component** and no snapshot files in the repo, so the proof was built:

1. A harness rendering all three components with a real asset-ref image, dumping `innerHTML`, captured before and after — `diff` reports identical.
2. A permanent regression test asserting the exact builder arguments per call site (e.g. `ImageCarousel` → width `[2400, 1200, 2400]`, quality `[85,85,85]`, fit `['max','max','max']`; `GalleryModal` → width `[1920, 192]`, height `[128]`). The `data:` cases assert the URI passes through, `srcSet` is absent, and the builder is not called at all.

5239 tests green; gallery story captured and inspected. Diff is confined to the src-resolution lines plus one new file, so #722's mosaic variant should merge trivially.

<h3>Greptile Summary</h3>

The PR adds a gallery image-source resolver that passes inline `data:` artwork through unchanged while preserving existing Sanity CDN transformations.
- Uses the resolver for main images and thumbnails in all three gallery components.
- Omits responsive `srcSet` values for inline artwork.
- Adds regression coverage for inline artwork and existing Sanity image transformations.

Inline artwork consistently bypasses the Sanity URL builder, responsive variants are omitted only for data URIs, and ordinary gallery assets retain their previous transformation parameters.

| Filename | Overview |
|----------|----------|
| src/lib/sanity/client.ts | Adds consistent detection and passthrough of inline image data while retaining the existing builder behavior for Sanity assets. |
| src/components/ImageCarousel.tsx | Resolves the primary image through the new helper and correctly suppresses srcSet for inline artwork. |
| src/components/GalleryModal.tsx | Applies the shared resolver to both the modal image and thumbnail rendering paths. |
| src/components/SimpleImageCarousel.tsx | Applies the shared resolver without changing the existing carousel loading or navigation behavior. |
| src/components/GalleryImageSrc.test.tsx | Covers inline data-URI passthrough and verifies that existing builder transformations remain unchanged across all affected components. |

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(gallery): render images whose artwor..."](https://github.com/cloudnativebergen/website/commit/3b830834475cc720d2436c814603c7e9a596435d) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=49828979)</sub>

### DIFF
--- a/src/components/GalleryImageSrc.test.tsx
+++ b/src/components/GalleryImageSrc.test.tsx
@@ -1,0 +1,162 @@
+/**
+ * @vitest-environment jsdom
+ */
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+import { render, cleanup } from '@testing-library/react'
+import { createImageUrlBuilder } from '@sanity/image-url'
+import { ImageCarousel } from '@/components/ImageCarousel'
+import { SimpleImageCarousel } from '@/components/SimpleImageCarousel'
+import { GalleryModal } from '@/components/GalleryModal'
+import type { GalleryImageWithSpeakers } from '@/lib/gallery/types'
+
+vi.mock('next-auth/react', () => ({
+  useSession: () => ({ data: null, status: 'unauthenticated' }),
+}))
+
+vi.mock('@/lib/trpc/client', () => ({
+  api: {
+    gallery: {
+      untagSelf: {
+        useMutation: () => ({
+          mutate: () => {},
+          mutateAsync: async () => {},
+          reset: () => {},
+          isPending: false,
+        }),
+      },
+    },
+  },
+}))
+
+// The image-url module is aliased to a chainable mock builder in
+// vitest.config.ts. client.ts builds one at import time, so every transform in
+// the app funnels through this single instance and its calls are inspectable.
+const MOCK_CDN_URL = 'https://cdn.sanity.io/images/mock/image.png'
+
+function builderCalls(method: 'width' | 'height' | 'quality' | 'fit') {
+  const builder = vi.mocked(createImageUrlBuilder).mock.results[0]
+    .value as Record<string, ReturnType<typeof vi.fn>>
+  return builder[method].mock.calls.map((call) => call[0])
+}
+
+const DATA_URI =
+  'data:image/svg+xml;base64,PHN2ZyB4bWxucz0iaHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmciLz4='
+
+const realImage: GalleryImageWithSpeakers = {
+  _id: 'image-1',
+  _rev: 'rev-1',
+  _createdAt: '2026-01-01T00:00:00Z',
+  _updatedAt: '2026-01-01T00:00:00Z',
+  photographer: 'Ada Lovelace',
+  date: '2026-01-01',
+  location: 'Bergen',
+  featured: true,
+  image: {
+    _type: 'image',
+    asset: { _ref: 'image-abc123-1200x800-jpg', _type: 'reference' },
+  },
+  speakers: [],
+  imageUrl: 'https://cdn.sanity.io/images/p/d/abc123-1200x800.jpg',
+  imageAlt: 'A real photo',
+}
+
+const generatedImage: GalleryImageWithSpeakers = {
+  ...realImage,
+  _id: 'image-2',
+  imageUrl: DATA_URI,
+  imageAlt: 'Generated artwork',
+}
+
+beforeEach(() => {
+  const builder = vi.mocked(createImageUrlBuilder).mock.results[0]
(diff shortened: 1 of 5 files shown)


Starefossen:
Three findings assessed in `af9255f2` — two real and fixed, one refuted.

**`isDataUri()` overbroad (copilot + qodo) — real.** `imageUrl` comes from `"imageUrl": image.asset->url`, i.e. dataset content, which in a multi-tenant install is organizer-authored and writable from Studio without passing the tRPC write path. Accepting any `data:` prefix and returning it verbatim into `` was too loose. Replaced with an anchored allowlist of raster image types, with a trailing `[;,]` anchor so prefix look-alikes like `data:image/pngx,` are rejected.

The repo's existing data-URI validators are all `data:image/…` shaped (invitation letters, signing, contract templates, and closest of all the sponsor logo rasteriser), so this matches the house pattern rather than inventing one.

**Deliberately its own list rather than reusing `RICH_TEXT_IMAGE_MIME_TYPES`**, for two reasons recorded in the code: dependency direction — the shared Sanity client must not import a homepage feature module — and semantics. That constant is an *upload* allowlist ("what may enter the asset pipeline"); this is a *render* allowlist ("what bytes we inline into a visitor's page from a string that never passed an upload gate"). Widening one should not silently widen the other. The comment names both so a reader finds the other.

**SVG excluded, aligned with the repo's existing position** rather than contradicting it: `SANITY_IMAGE_REF_PATTERN`'s comment states SVG is excluded from image asset ids because "an SVG is an active document", and the SVG this app does accept is only ever stored after `sanitizeSvgUpload` and re-checked at render. A `data:` URI in an `` passes through neither and cannot. Not a live script hole — `` renders SVG in secure static mode — but accepting it would open a second, unsanitized SVG intake behind the sanitizer's back.

**Thumbnails bypass (qodo) — real, though not at the cited line.** The `srcSet` ternary was already correct; the bypass was the **thumbnail strip**, which built src and srcSet straight from the CDN builder. A data-URI image would have rendered in the main frame and 404'd in its own thumbnail immediately below it. Now routed through the same helper. Every `` in all three components was checked; no others remain.

**`vi.mock('@/lib/trpc/client')` (qodo) — refuted.** AGENTS.md says "mock at boundaries", and that module *is* the tRPC HTTP boundary — importing it for real builds a React Query client and issues network calls. 15+ sibling tests mock it identically, including the nearest neighbour in the same directory. No rule file forbids it. Kept, with a comment recording the justification and pointing at the sibling.

**A rejected data URI falls through to the CDN builder** — exactly the pre-PR behaviour — so it never reaches the DOM. Real-image assertions still hold verbatim, and there are now three `it.each` suites over 7 rejected URIs plus a thumbnail regression test locking the exact builder arguments. 31 tests in the file, up from 9.

**Flagged, not changed:** three other gallery renderers (`PhotoBentoGrid`, admin `ImageGrid`, `PhotoGalleryBuilder`) still build from the CDN builder and would show the same 404 for a data-URI image — the last of those returns a raw `imageUrl` from a `catch`, an unguarded pass-through of the same kind. Out of scope here.

PR #727, opened 2026-08-03
feat(email): brand every email to the conference sending it

Starefossen:
A conference could configure its brand colours and its **email still went out in Cloud Native blue**. For an external customer that means their speakers, sponsors and attendees receive mail branded as somebody else's conference.

## The audit, confirmed line by line

- Exactly **2 of 24** styled send paths passed a brand colour (`speaker/ticket-email.ts:58`, `events/handlers/emailNotification.ts:72`) — and even those hard-coded `#1D4ED8` for titles and left borders.
- `EmailButton` defaulted to `#1D4ED8` with a matching shadow, across 11 call sites, **none** of which passed a colour. Every button in every email ever sent was house blue.
- `portableTextToHTML.tsx` hard-coded the brand blue 8 times (links, strong, h1–h6) with no colour parameter at all, and it feeds broadcasts, sponsor-CRM mail and contract email.

**Now 24 of 24 branded.** The two remaining paths carry no colour at all (one unstyled, one plaintext).

## Unthemed output is byte-identical, and the proof came first

21 snapshots covering every template, the badge HTML and `portableTextToHTML` were generated **in the first commit, on the untouched tree** — a genuine pre-change baseline rather than a post-hoc one. They pass unmodified after both feature commits, and `git log -p` shows the snapshot file untouched since creation.

The mechanism that makes that hold rather than luck: an unthemed conference resolves to the frozen house palette, and every literal swap goes through helpers that return the **original literal verbatim** when the palette is the default. That is what allowed four disagreeing house blues (`#1D4ED8`, `#1E40AF`, `#2563EB`, `#0284C7`) to collapse onto the tenant accent without changing a single unthemed byte — a plain "delete the literals and inherit" would have broken it.

A gate test now fails any future email file that pins a house brand hex outside those helpers.

## Structure

**Inherit once** — a brand context resolved in `BaseEmailTemplate`; `EmailButton` and `EmailSectionHeader` read it, so a new template is branded by doing nothing. Titles got a `titleTone: 'brand' | 'neutral'` rather than a defaulted colour, because the templates genuinely disagree — half coloured, half slate — and collapsing them would have changed unthemed mail.

**Thread once** — `renderEmailTemplate` and `convertPortableTextToHTML` resolve the colour themselves, branding four broadcast-shaped senders through one signature.

**Widen the theme-blind types** — 7 found, including three the design missed and one hand-built projection on the discount page that dropped `theme` on the floor.

## Contrast

White-on-brand buttons need to stay legible for a pale tenant colour. `ensureContrastWithWhite` bisects the OKLab **lightness** axis until WCAG AA 4.5:1 is met, hue preserved — so `#FACC15` becomes a dark gold that is still recognisably yellow, not grey. A colour that already passes is returned verbatim **including casing**, which is part of why byte-identity survives.

## Status colours are exempt, and asserted

Reject red, waitlist orange/amber, success green all stay literal. A red error in a conference's brand colour would be worse than a house-coloured one.

## One residual, flagged rather than guessed

`EmailButton variant="secondary"` fills `#6366F1` (indigo, one call site). It is not one of the house blues, so branding it would change unthemed bytes — and it needs a decision on whether "secondary" is brand or neutral chrome. Left alone deliberately.

Note this branch was built on `main`, so it uses `emailBrandColor`; #710 introduces `resolveBrandPair` and the two will want reconciling once that lands. 5344 tests green.

<h3>Greptile Summary</h3>

This PR introduces tenant-aware branding throughout the outbound email system.
- Adds a derived email palette with contrast-aware accent colors and default-palette compatibility.
- Propagates conference themes through email templates, rich-text conversion, sponsor, speaker, signing, workshop, volunteer, messaging, and reminder flows.
- Adds shared brand context for email primitives plus snapshot, themed-rendering, gate, and Storybook coverage.

The PR appears safe to merge, with a non-blocking visual inconsistency in themed secondary email buttons.

Tenant branding is propagated consistently through the main rendering paths, while the remaining issue is limited to the secondary button retaining an indigo fill but receiving a tenant-colored shadow.

**Files Needing Attention:** src/components/email/EmailComponents.tsx

| Filename | Overview |
|----------|----------|
| src/lib/branding/email.ts | Adds centralized, validated email palette derivation while preserving the legacy palette for unthemed conferences. |
| src/lib/branding/color.ts | Adds WCAG contrast calculations and OKLab lightness adjustment for white-on-brand email content. |
| src/components/email/BaseEmailTemplate.tsx | Resolves tenant branding once and publishes it to nested email primitives through context. |
| src/components/email/EmailComponents.tsx | Makes primary buttons and section headers inherit branding, but secondary buttons now combine a fixed indigo fill with a tenant-colored shadow. |
| src/lib/email/portableTextToHTML.tsx | Applies the resolved brand palette to links, emphasis, and rich-text headings. |
| src/lib/email/route-helpers.ts | Centralizes branding for broadcast-style email shells and Portable Text bodies. |
| src/server/routers/sponsor.ts | Threads conference branding through sponsor email paths and mirrors branded discount content between preview and send output. |
| src/lib/email/workshop.ts | Applies conference branding to the independently rendered workshop email variants. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart LR
  Theme[Conference theme] --> Resolve[emailBrandColor]
  Resolve --> Palette[Email brand palette]
  Palette --> Base[BaseEmailTemplate provider]
  Base --> Primitives[Buttons and section headers]
  Resolve --> RichText[Portable Text HTML]
  Resolve --> Direct[Raw and specialized email templates]
  Primitives --> Email[Rendered outbound email]
  RichText --> Email
  Direct --> Email
```

<sub>Reviews (1): Last reviewed commit: ["feat(email): brand the remaining 22 send..."](https://github.com/cloudnativebergen/website/commit/46c64d8f44be45fbd14d867ac9d473705fd08b3e) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=49834435)</sub>

- Knowledge Base — [Workshop management and volunteer sign-up](https://app.greptile.com/cloudnativedays/-/custom-context/knowledge-base/cloudnativebergen/website/-/docs/workshop-volunteer.md)
</details>


Starefossen:
Five findings assessed in `fb8e3612`. **All 21 unthemed byte-identity snapshots still pass unmodified** — `git diff` against the baseline commit on the snapshot file is empty.

**Secondary button shadow (greptile + qodo) — real, but it IS the known `#6366F1` residual, so not decided unilaterally.**

Evidence it is inherited rather than introduced here: before this PR the line was one fixed house-blue shadow under *both* variants, so indigo-fill-over-blue-shadow already existed — branding only changed which colour the mismatch is. Applying greptile's exact suggested patch fails **2 of the 21 baseline snapshots**; qodo's variant is the same edit. The one byte-safe alternative — pinning the house-blue shadow for secondary — silently decides that "secondary" means neutral chrome, which is the product question. A comment now records why it is unfixed so it is not re-raised. One production call site.

**Mutable default palette (qodo) — real, fixed, and it was the worst one to leave open.** `DEFAULT_EMAIL_BRAND_PALETTE` was the only unfrozen palette while every derived one was frozen — and derived palettes are cached per hex, whereas the default is a module singleton handed to *every* unthemed send. One mutation would have re-branded every subsequent email in the process. Now frozen, and all palette fields are `readonly` so the compiler rejects mutating derived ones too.

**Gate test (qodo) — real, and it had two more holes than reported.** This is the test that stops a future email file pinning a house brand hex outside the helpers, so a weak version is worse than none:
1. Path separators — now normalised, so exemptions match on Windows.
2. *Stale exemptions* — a renamed or deleted exempt file left a dead entry that the next file at that path would inherit for free. Now asserted against the files the walk actually finds.
3. **The gate was blind to the exact literal it exists to catch.** `rgba(29, 78, 216, 0.25)` contains no `#1d4ed8`, so every box-shadow in the codebase was invisible to it. It now matches `rgb()`/`rgba()` spellings — verified with a throwaway probe the old gate passed and the new one fails.
4. The house blue now comes from `DEFAULT_PRIMARY_COLOR` rather than a hard-coded copy, and an email directory the gate was not scanning is now covered.

**Missing barrel (qodo) — real convention gap, deliberately out of scope.** `src/lib/branding` predates this PR and is imported deeply from 81 files / 96 sites; only 17 of those are this PR's. Only 16 of 67 `src/lib/*` subdirectories have barrels, so this is a repo-wide pattern and the flagged line matches every other consumer. Adding it means rewriting 81 files and pulling four modules into one entry point for client components — a separate refactor, not a rider on an email change.

5347 tests green.

Starefossen:
**The `href` interpolation finding is fixed** in `af7c2d81`, and it was a real one — thank you.

The link mark interpolated `value.href` straight into an `href="…"` attribute of a hand-built HTML string, with no scheme check and no escaping. Both halves mattered: a stored `javascript:` href rode out in mail a client may open in a browser context, and a value containing a quote closed the attribute so everything after it became markup nobody wrote.

It now goes through `toSafeRichTextHref` — the same predicate the site's rich text and the Studio use, so an href the page refuses cannot be smuggled out through email instead — and is escaped for attribute context.

**Worth recording how the tests were checked**, because the first version of one of them was wrong in an instructive way. A naive `expect(html).not.toMatch(/onmouseover=/)` **passed against the vulnerable code**: once the quote is escaped, the text `onmouseover=` is still present, harmlessly, inside the href value. The assertion now counts raw quotes in the anchor instead. Verified by reverting the fix — 2 of the 4 tests fail without it.

The other three threads were assessed earlier: the two on `EmailComponents` are the known `#6366F1` secondary-button residual (it predates this PR, and branding it would change unthemed bytes — left for a product decision), and the deep-import one is a repo-wide pattern across 81 files that belongs in its own refactor.

6094 tests green, lint 0 errors, production build passing.

PR #728, opened 2026-08-03
fix(security): close cross-tenant leaks before a second customer exists

### LINKED ISSUE #723: SECURITY: organizer authz gate falls back to the GLOBAL organizer set across all tenants
Found during the multi-tenant readiness audit. **Not exploitable today** (one organization exists) — becomes a cross-tenant privilege escalation the moment a second customer's data is in the dataset, so it must be closed BEFORE onboarding customer #1.

## The hole

`getOrganizerSpeakerIds()` (`src/lib/notification/sanity.ts:744-746`) resolves its scope like this:

```ts
const organizerScope = resolvedOrgId
  ? `*[_type == "conference" && organization._ref == $orgId].organizers[]._ref`
  : `*[_type == "conference"].organizers[]._ref`
```

When the org cannot be resolved it returns **every organizer of every conference**. That is a deliberate legacy bridge for recipient selection, and it warns — but it is also used as an **authorization gate**:

- `src/server/routers/workshop.ts:118` — manage announcements
- `src/server/routers/workshop.ts:223` — send announcements

Both call it with **no argument**, so `orgId` is `undefined`, the org never resolves, and the global branch is taken every time. Any organizer of any conference passes the gate for any workshop.

## Failure scenario

Two customers on the platform. An organizer of conference A calls `workshop.sendAnnouncement` for a workshop belonging to conference B. `isOwner` is false; `isOrganizer` is true because A's organizer id is in the global set. B's workshop attendees receive an announcement written by someone from another organization.

## Why it survived

The tenancy lint rule cannot see this: the filter is interpolated, not a literal, and the call sites pass no org at all rather than passing `null`. This is the same class as the `scopedFetch({orgId: null})` fail-open footgun — an unresolvable tenant yields GLOBAL data instead of nothing.

## Fix direction

Fail closed. An authz gate must never fall back to a global set — if the org cannot be resolved, deny. The recipient-selection use may legitimately need the bridge; the two authz call sites do not, and should pass the workshop's own conference/org and get an empty set on failure. Consider splitting the function so an authz caller cannot reach the global branch at all.

Related: #683, #687, #688 (multi-tenant auth), and the fail-open scoping class generally.

Starefossen:
Closes #723. Found by the multi-tenant readiness audit. **Nothing here is exploitable today** — one organization exists — but every one becomes live the moment a second paying customer's data is in the shared dataset.

## 1. Organizer authz gate (#723) — confirmed, and wider than reported

`getOrganizerSpeakerIds()` fell back to `*[_type == "conference"].organizers[]._ref` — every organizer of every conference — and both workshop gates called it with **no arguments**, so the global branch was taken every time. An organizer of conference A could manage and send announcements to conference B's workshop attendees.

Both gates now use the already-fail-closed `isOrganizerForCurrentOrg`. More importantly the trap is removed **structurally**: the global branch is now a private function, and the three exports are explicit — current-org (empty when unresolvable), for-a-named-org (empty on null), and one clearly-named cross-org read used only by the nudge cron. No exported path reaches global by omission, so the next caller cannot repeat this by writing zero arguments.

## 2. Gallery reads — confirmed

Two defects: no id meant the whole dataset, and `!defined(conference)` handed every conference-less image to every tenant. An unknown host returned **all tenants' featured images plus 50 more**. Scope is now required and its predicate unconditional; a blank scope returns empty **without querying at all**.

`listMine`/`countMine` deliberately scope to the ORG rather than the conference, so a CND speaker still sees their photos across all three editions.

## 3. Shared-speaker proposals — confirmed, plus a call site the audit missed

The conference filter fell back to `''`. The audit named the badge page; `/admin/speakers` had the identical hole.

## Beyond the brief: cross-tenant gallery WRITES

Not in the audit, and a different and more serious class — missing ownership check rather than fail-open read. `gallery.admin.update/delete/toggleFeatured` and `untagSelf` took an image id straight from client input with no tenant check, so an organizer of A could **edit or delete B's photos** by id. Closed, including a guard that a patch may not reassign an image to another conference. Happy to split it out if you'd rather review it separately.

## Root enablers

**The lint rule was strengthened and pays out immediately.** `no-unscoped-groq` now sees interpolated filters (`*[${filter}]` — previously invisible, and the exact shape of two of these bugs), optional tenant filters, and `scopedFetch({orgId: null})`. It flags 9 real sites including the ones the audit named. Verified no coverage regression: the `unscoped` count moved 241 → 234 and every drop is duplicate collapse within a single template.

**`scopedFetch` now throws on an empty scope** rather than degrading to a global read — safe to do now because it has zero production call sites, so the trap is gone before it acquires users.

**`getConferenceForDomain`'s truthy `{}` is NOT fixed** — a ~100-call-site refactor judged too large to ride along with a security fix. Two dead `if (!conference)` checks on touched paths were replaced with a real `isUnknownHost` guard; **~13 remain** elsewhere and are tracked as their own work item.

## One check needed before merge

Dropping `!defined(conference)` means a gallery image with no conference reference is now returned to nobody. `createGalleryImage` has always required one, so only hand-made Studio documents could be in that state — but this could not be confirmed without a dataset token:

```groq
count(*[_type == "imageGallery" && !defined(conference)])
```

If that is non-zero, those images need a conference assigned or CND loses them from its public gallery.

## Not fixed, named deliberately

`src/lib/messaging/standing.ts` holds the same global-organizer bridge, gating who an organizer may name as a thread subject. Left as its own review surface and documented in `docs/ORGANIZATION_TIER.md`. `scripts/**` stays lint-allowlisted despite running with the write token — cross-tenant by design, needs per-script annotations first.

5281 tests green, production build exit 0.

<h3>Greptile Summary</h3>

This PR closes several fail-open tenant-scoping paths.
- Replaces workshop organizer-set authorization with current-organization authorization.
- Requires gallery read scopes and adds ownership checks to gallery mutations.
- Scopes shared-speaker proposal projections and organizer notification routing.
- Makes scopedFetch reject empty scopes and strengthens the GROQ lint rule.

The PR appears safe to merge, with a non-blocking recommendation to make gallery ownership checks revision-bound so concurrent reassignment cannot invalidate authorization before mutation.

The changed paths consistently fail closed and enforce tenant scopes, but gallery mutations still rely on a separate ownership read followed by an unconditional ID-based write.

**Files Needing Attention:** src/server/routers/gallery.ts, src/lib/gallery/sanity.ts

The gallery ownership checks substantially improve isolation, but their read-before-write design leaves a narrow concurrent reassignment window because the final mutation is not revision-bound.
</details>

| Filename | Overview |
|----------|----------|
| src/server/routers/gallery.ts | Adds tenant-scoped reads and mutation ownership guards; the guards are not atomic with their subsequent writes. |
| src/lib/gallery/sanity.ts | Requires explicit gallery scopes and adds tenant lookup support, while existing ID-based mutations remain unconditional. |
| src/server/routers/workshop.ts | Replaces global organizer-set checks with the shared current-organization authorization gate. |
| src/lib/notification/sanity.ts | Splits current-org, named-org, and explicitly cross-org organizer reads into fail-closed APIs. |
| src/lib/sanity/scoped.ts | Makes scopedFetch throw before querying when no tenant scope is available. |
| src/lib/speaker/sanity.ts | Prevents cross-organization proposal projections by falling back to conference scope when organization scope is unavailable. |
| eslint-rules/no-unscoped-groq.js | Expands lint detection to interpolated, optional, and explicitly null tenant-scope patterns. |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
  participant A as Tenant A request
  participant G as Gallery router
  participant S as Sanity
  participant E as Concurrent editor
  A->>G: update/delete image ID
  G->>S: read image tenant
  S-->>G: tenant A
  E->>S: reassign image to tenant B
  G->>S: unconditional patch/delete by ID
  S-->>G: mutation succeeds on tenant B document
```

<sub>Reviews (1): Last reviewed commit: ["fix(tenancy): fail closed on an unresolv..."](https://github.com/cloudnativebergen/website/commit/b27e54da0f866876606bb2d6bff1059d9bd8e133) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=49837372)</sub>

**Context used:**

### DIFF
--- a/__tests__/lib/messaging/notify.test.ts
+++ b/__tests__/lib/messaging/notify.test.ts
@@ -18,7 +18,8 @@ import { describe, it, expect, vi, beforeEach } from 'vitest'
 
 vi.mock('@/lib/notification/sanity', () => ({
   upsertMessageNotifications: vi.fn(async () => {}),
-  getOrganizerSpeakerIds: vi.fn(async () => ['org-1']),
+  // The fan-out reads the CONFERENCE's org explicitly (#723).
+  getOrganizerSpeakerIdsForOrg: vi.fn(async () => ['org-1']),
 }))
 
 vi.mock('@/lib/sanity/client', () => ({
@@ -44,7 +45,7 @@ vi.mock('@/lib/messaging/sanity', async (importActual) => {
 
 import {
   upsertMessageNotifications,
-  getOrganizerSpeakerIds,
+  getOrganizerSpeakerIdsForOrg,
 } from '@/lib/notification/sanity'
 import { clientReadUncached } from '@/lib/sanity/client'
 import { getConversationPreferencesFor } from '@/lib/messaging/sanity'
@@ -57,7 +58,7 @@ import type { Conference } from '@/lib/conference/types'
 
 type LooseMock = ReturnType<typeof vi.fn>
 const upsertMock = upsertMessageNotifications as unknown as LooseMock
-const organizersMock = getOrganizerSpeakerIds as unknown as LooseMock
+const organizersMock = getOrganizerSpeakerIdsForOrg as unknown as LooseMock
 const fetchMock = (clientReadUncached as unknown as { fetch: LooseMock }).fetch
 const prefsMock = getConversationPreferencesFor as unknown as LooseMock
 const emailMock = sendMessageEmails as unknown as LooseMock
--- a/__tests__/lib/messaging/nudge.test.ts
+++ b/__tests__/lib/messaging/nudge.test.ts
@@ -12,7 +12,9 @@
 import { describe, it, expect, vi, beforeEach } from 'vitest'
 
 vi.mock('@/lib/notification/sanity', () => ({
-  getOrganizerSpeakerIds: vi.fn(async () => ['org-1', 'org-2']),
+  getOrganizerSpeakerIdsForOrg: vi.fn(async () => ['org-1', 'org-2']),
+  // The candidacy superset is now an explicitly named cross-org read (#723).
+  getAllOrganizerSpeakerIdsAcrossOrgs: vi.fn(async () => ['org-1', 'org-2']),
   createNotifications: vi.fn(async () => {}),
 }))
 
@@ -23,7 +25,7 @@ vi.mock('@/lib/sanity/client', () => ({
 
 import { clientReadUncached, clientWrite } from '@/lib/sanity/client'
 import {
-  getOrganizerSpeakerIds,
+  getOrganizerSpeakerIdsForOrg,
   createNotifications,
 } from '@/lib/notification/sanity'
 import {
@@ -82,7 +84,7 @@ beforeEach(() => {
   vi.spyOn(console, 'error').mockImplementation(() => {})
   vi.spyOn(console, 'log').mockImplementation(() => {})
   vi.spyOn(console, 'warn').mockImplementation(() => {})
-  vi.mocked(getOrganizerSpeakerIds).mockResolvedValue(['org-1', 'org-2'])
+  vi.mocked(getOrganizerSpeakerIdsForOrg).mockResolvedValue(['org-1', 'org-2'])
   // B4: the loop now batch-resolves each conversation's conference → owning org
   // before scoping recipients. That extra read is the ONLY fetch carrying
   // `organization._ref`; route it to a stable mapping so conf-1 owns org-1 (the
@@ -187,7 +189,7 @@ describe('routing + stamping', () => {
   })
 
   it('skips (does not stamp) an unassigned thread when there are no organizers', async () => {
-    vi.mocked(getOrganizerSpeakerIds).mockResolvedValue([])
+    vi.mocked(getOrganizerSpeakerIdsForOrg).mockResolvedValue([])
     installPatch()
     readMock.fetch.mockResolvedValueOnce([{ ...unassignedProposalConv }])
 
--- a/__tests__/lib/notification/sanity.test.ts
+++ b/__tests__/lib/notification/sanity.test.ts
@@ -35,9 +35,9 @@ vi.mock('@/lib/push/send', () => ({
   sendPushForNotifications: vi.fn(async () => {}),
 }))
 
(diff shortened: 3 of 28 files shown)


Starefossen:
**Pre-merge dataset check: clear.**

```groq
count(*[_type == "imageGallery" && !defined(conference)])
```

returned **0** against the production dataset. So dropping the `!defined(conference)` predicate strands nothing — there are no conference-less gallery images to lose, and CND's public gallery is unaffected.

That was the only unknown in this PR that could not be settled from the code. Adversarial review in progress before merge.

Starefossen:
Both threads assessed.

**`NOT_FOUND` masked as a 500 (qodo) — real, fixed.** `requireCurrentOrgId` throws `NOT_FOUND` on an unresolvable host, and that *is* the fail-closed signal this PR introduces — but `listMine`/`countMine` wrapped every error as `INTERNAL_SERVER_ERROR`, hiding a tenancy failure behind a generic 500 in the client and, more importantly, in monitoring. A security change whose signal is invisible to monitoring is half a change. Both now rethrow `TRPCError` first, matching what the admin paths in the same router already do.

**Non-atomic ownership check (greptile, P2 security) — real, split out as #729.** The tenant read and the mutation are separate operations with no transaction or revision condition, so a concurrent reassignment could move the document between them.

Assessed as hardening rather than a live hole: exploiting it requires an actor who can already reassign an image's `conference` reference — that is dataset write access via Studio or an API token, and anyone holding it can damage the document directly. It is also not exploitable at all today, with one organization.

Split out rather than fixed here for two reasons. Bundling a concurrency fix into a security PR that had already outgrown its brief (it picked up the cross-tenant *write* class along the way) would make the isolation changes harder to review. And the fix needs a decision, not a patch: revision-conditioned patches cover three paths but Sanity's delete has no equivalent precondition, so delete needs a query-constrained mutation regardless — and doing that properly rewrites four helpers with other callers. #729 lays out the options.

**Pre-merge dataset check came back 0**, so dropping `!defined(conference)` strands nothing.

5281 tests green, production build exit 0. Merging on the owner's approval; an independent adversarial review of this branch is still running and anything it finds will land as a follow-up rather than being lost.

PR #733, opened 2026-08-04
perf(frontpage): the hygiene batch a measured audit asked for

Starefossen:
A measured audit concluded the recent customisation work cost essentially nothing (+5 KB gz, identical LCP/CLS/TTFB, zero warm Sanity queries) but found real pre-existing costs. This is those, with every claim re-measured — the audit turned out to be wrong twice, in opposite directions.

## 1. OG fonts were re-fetched on every request

`loadBrandFonts` fetched three TTFs per call with no memo. Measured against a local server serving the real files: **3 requests / 789 KB / 3.98 ms per render**. Now a module-level promise cache mirroring the existing `loadBadgeFontFiles` contract — failures not cached, reset hook for tests.

Measured over 20 renders: **60 requests / 15,778 KB → 3 requests / 789 KB**, 0 after the first on a warm instance.

Deliberately **not** keyed by domain: `domain` only picks the origin and `/public/fonts` is identical on every host, so keying it would multiply ~808 KB of resident buffers by tenant count. It also now throws on a non-ok response — previously a 404 body was handed to satori as font bytes.

Serverless caveat, stated plainly: module state is per-instance, so a cold start still pays one fetch. That is the ceiling without bundling fonts as function assets, and it matches what `lib/badge/fonts.ts` already ships. The per-*request* refetch is gone.

## 2. Featured-gallery fetch was bounded at 1000

`getFeaturedGalleryImages` fell back to `limit || 1000` while `GALLERY_CONSTANTS.LIMITS.FEATURED_IMAGES = 8` sat dead. Now uses the constant. Also removed the asymmetry in `getConferenceForDomain` where only one branch defaulted the limit — that asymmetry is precisely what let the homepage through.

Output-identical today: the live homepage has **zero** featured gallery images, so no tenant is near the bound.

## 3.  — bigger win than claimed, and NOT literally pixel-identical

Measured against the live CDN with a real Chrome `Accept` header:

| asset | before | after |
|---|---|---|
| speaker portrait 640×800 | 54.4 KB jpeg | **31.9 KB avif (−41%)** |
| second portrait | 39.5 KB jpeg | 16.7 KB avif (−58%) |
| gallery frame 1200/q85 | 83.3 KB jpeg | 29.1 KB avif (−65%) |
| PNG headshot 640×800 | 773.7 KB png | **26.6 KB webp (−97%)** |

The audit's "55.7 → 45.0 KB, −19%" measured webp; Chrome gets AVIF.

**But "zero pixel change" does not survive measurement.** Decoded pixels differ — PSNR 35–42 dB, ~55% of channels off by ≥1. Measured against a *lossless* `fm=png` rendition of the identical transform, however, it is **fidelity-neutral**: 37.38 vs 35.80, 38.38 vs 38.49, 42.07 vs 42.35 dB. Both are lossy renditions of one master, so this is a codec swap rather than a quality reduction — and in one case the new rendition is closer to ground truth.

Applied on that basis, but it is in its own commit (`190a7cb4`) so it can be reverted alone if the pixel-identity constraint is read strictly. **Worth a human eye on one real photo.**

## 4. Reduced motion — an accessibility defect, not just a performance one

**The background pattern ignored `prefers-reduced-motion` entirely**: with the preference emulated, **98 of 98** nodes still ran their infinite float. The cause was that the animation shorthand was an *inline* style, which no stylesheet can override. Moved the seeded duration and delay to custom properties with a `prefers-reduced-motion` override — the same shape as the existing `.animate-marquee-dynamic` pair.

After: `reduce` → **0** animating nodes; `no-preference` → still 98. The default path was verified by dumping computed animation, geometry, opacity, z-index and filter for all 98 nodes before and after — identical, the only difference being the live `transform` sample, as it must be for a running animation.

**The carousel autoplayed offscreen.** Now gated on reduced-motion, viewport intersection and page visibility.

One real find: the obvious effect-based version of that hook forced a second render of every consumer on mount, making each carousel build its image URLs twice and breaking 11 existing assertions. Deciding the no-`IntersectionObserver` fallback in the state initialiser fixes it, and the hook documents why.

## Verification

typecheck clean, lint 0 errors, **5375 tests green**, production build succeeds, client chunks **2157.0 → 2157.7 KB gz (+0.03%)** on clean builds of baseline vs branch.

Not done, deliberately: the 197 KB logo SVG inlined four times (~2.3 MB of the 2.9 MB live page). That needs an upload-time SVGO pass and single-instance rendering — its own PR with its own risk.

<h3>Greptile Summary</h3>

The PR reduces front-page and OG rendering costs while improving motion accessibility.
- Caches self-hosted OG font fetches and rejects unsuccessful responses.
- Bounds featured-gallery reads using shared gallery constants.
- Enables browser-negotiated Sanity image formats while preserving server-rasterizer behavior.
- Disables seeded background animation for reduced-motion users and gates carousel autoplay by motion preference, viewport intersection, and page visibility.

The PR appears safe to merge, with the non-blocking caveat that the changed carousel behavior lacks the repository-required Storybook visual validation.

The changed runtime paths are bounded and lifecycle-consistent, while the remaining concern is limited to missing isolated visual coverage for browser-dependent UI states.

**Files Needing Attention:** src/components/ImageCarousel.tsx, src/hooks/useIsOnScreen.ts

| Filename | Overview |
|----------|----------|
| src/lib/og/helpers.ts | Adds a failure-resetting module-level promise cache for the three deployment-owned OG font files. |
| src/lib/conference/sanity.ts | Normalizes gallery defaults and delegates omitted featured limits to the gallery data-access contract. |
| src/lib/gallery/sanity.ts | Replaces the effectively unbounded featured-image fallback with the configured eight-image default. |
| src/lib/sanity/client.ts | Adds browser-oriented automatic image-format negotiation with an explicit opt-out for server rasterizers. |
| src/components/ImageCarousel.tsx | Gates autoplay by intersection and reduced-motion state, with unit coverage but no required Storybook visual fixture. |
| src/hooks/useIsOnScreen.ts | Adds a lifecycle-clean observer hook combining viewport intersection with document visibility. |
| src/components/CloudNativePattern.tsx | Moves seeded animation timing into CSS variables so reduced-motion styles can disable animation. |
| src/styles/tailwind.css | Defines seeded float animation and a reduced-motion override. |

</details>

<sub>Reviews (1): Last reviewed commit: ["a11y: honour prefers-reduced-motion in t..."](https://github.com/cloudnativebergen/website/commit/81b76e1946b5048a662c07403c81308374c05526) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=49941854)</sub>

**Context used:**


Starefossen:
**One real finding fixed, one false positive, one suggestion deferred.**

### Fixed: the gallery was read TWICE on every render (`2e218e5b`)

Good catch, and it was a genuine defect in a PR whose entire purpose is to remove wasted work.

The `Promise.all` held three promises while the destructuring bound two. The refactor was meant to *replace* the hardcoded `limit: 50` with `GALLERY_CONSTANTS.LIMITS.DEFAULT_GALLERY_LIMIT`; instead it added the new call alongside the old one, so every non-featured render issued the gallery query twice and discarded one result. The stale call is now gone.

One correction to the report: the constant **is** 50, so the two calls were behaviourally identical and the surviving binding returned the same data either way. The harm was the duplicate query, not a wrong limit.

That detail is also why this needed a test rather than more review attention — nothing rendered differently, so no assertion about output could ever have caught it. `__tests__/lib/conference/gallery-fetch.test.ts` asserts the **call count**, plus that the shared constant is the source of the limit and that `featuredOnly` reads nothing extra. Restoring the duplicate fails the count test; restoring it does not fail the limit test, and the comment there says so rather than implying coverage it does not have.

### Not a bug: `{ ...(null) }`

Object spread ignores `null` and `undefined` — only *iterable* spread throws:

```js
const animated = false
{ a: 1, ...(animated ? { b: 2 } : null), c: 3 }  // → { a: 1, c: 3 }
[...null]                                        // → TypeError: null is not iterable
```

Verified on this repo's Node. Per spec (`CopyDataProperties`), a `null`/`undefined` source is a no-op. No change made.

### Deferred: carousel visual fixture

Fair, but the intersection/visibility/reduced-motion behaviour wants a story with a controllable viewport and a motion preference, which is a fixture of its own rather than a line in this batch. Not blocking the measured fetch-and-render wins here.

---

Full suite green (5988), `tsc --noEmit` clean, lint 0 errors, formatting clean.

PR #739, opened 2026-08-04
feat(admin): the front-page composer's live preview

Starefossen:
The front-page composer's live preview: the real page, rendering the real section components, updating as the organizer edits — instead of the structural band list that today's editor shows and that the code itself describes as "NOT a page render".

This is the **preview surface**. The split-pane workspace that mounts it is a separate PR.

## Fidelity

It renders the actual section components, wrapped in the real `Header` and `Footer` and the tenant theme, so what the organizer sees is what ships. That is possible because the homepage render tree has zero server-only imports — verified during design, and the reason a client-rendered preview is the genuine article rather than an approximation.

Route: `/admin/homepage-preview`, in a sibling `(preview)` route group repeating the same organizer gate. **The design was wrong on one point of fact** and this corrects it: it claimed the route could inherit the admin auth layout, but `AdminLayout` renders the sidebar, command palette and notification chrome, and a page cannot opt out of an ancestor layout — so framing it would have put the admin shell around the tenant's front page.

## Three things that would have made it worse than useless

**It must not fire real analytics.** Pirsch's script loads from the ROOT layout, so it is present on admin routes — a preview rendering real CTAs would have quietly corrupted the conference's own click statistics every time someone edited their front page. A MutationObserver strips the tracking attributes across the preview subtree, not just once on mount: a carousel step or accordion toggle re-renders with no state message behind it and would otherwise leave live handles.

Verified three ways rather than assumed: a jsdom test asserting zero tracking attributes across all 13 section types (with an anti-vacuity probe proving the sweep has something to strip), every anchor dispatching a real click and reporting `defaultPrevented`, and a real Chromium run reporting zero.

**A throwing section must not take down the editor.** Per-band error boundaries that reset on the next state push — and deliberately do *not* retry on an unrelated re-render, which would loop the boundary on every pointer move.

**A new conference's preview would otherwise be blank.** Several bands self-hide when empty, which is exactly when an organizer most needs to see what they are designing. Design mode fills only empty collections with sample content, marked with amber chips, and ghosts bands that will self-hide on the live site. Live mode shows exactly what visitors get, with the rail explaining any band that vanished.

## Mechanism

Versioned, origin-checked `postMessage` in one module. Guards are structural rather than tag-based, and a state push is rejected as *outbound* so a hostile frame cannot drive the composer. Until the parent speaks, the saved composition renders.

Viewport truth: mobile is a real 390px frame; desktop is the 1440px layout scaled via `transform` on the iframe *element*, which does not change the iframe's layout viewport.

The preview data query is deliberately uncached and required threading an `uncached` option through `getConferenceForDomain` — the public page's `'use cache'` path is unaffected, which a later performance audit independently confirmed by measuring zero Sanity queries on warm public requests.

## One trap worth recording

Placeholder gallery tiles carry a deliberately dead asset ref, because the image builder throws on a malformed one. The CDN URL spells the asset differently from the ref, so the guard matches on the 40-character hash rather than the raw `_ref`. A story handler renders a loud red failure plate if that swap ever breaks, so a regression cannot be quiet. Browser result: zero CDN requests issued.

5359 tests green, production build exit 0 with the route present.

<h3>Greptile Summary</h3>

The PR adds an organizer-only, chrome-free homepage preview route that renders the production homepage components from unsaved postMessage state.
- Adds design and live preview modes with placeholder content, per-band status chrome, and error isolation.
- Introduces a versioned preview messaging protocol and DOM containment for navigation, analytics, theme changes, and placeholder images.
- Adds an uncached, host-scoped admin query matching the public homepage data dependencies.

The placeholder image request race should be fixed before merging because the current post-commit guard cannot guarantee that fake Sanity assets never reach the CDN.

Design-mode gallery images are initially committed with eagerly loaded CDN URLs, while substitution occurs later in a passive effect, leaving a concrete window for 404 requests and broken-image flashes.

**Files Needing Attention:** src/components/admin/preview/usePreviewDomGuard.ts

| Filename | Overview |
|----------|----------|
| src/app/(preview)/admin/homepage-preview/page.tsx | Adds the authenticated preview document, state-message handling, theme synchronization, and size reporting. |
| src/components/admin/preview/HomepagePreview.tsx | Renders real homepage components in design or live mode with placeholders and per-band isolation. |
| src/components/admin/preview/usePreviewDomGuard.ts | Strips analytics and replaces placeholder image URLs, but replacement occurs after CDN-backed image elements are committed. |
| src/lib/homepage/previewProtocol.ts | Defines versioned message envelopes, constructors, and structural guards for preview communication. |
| src/lib/homepage/placeholders.ts | Generates deterministic, marked sample content and inline artwork for otherwise empty homepage bands. |
| src/server/routers/conference.ts | Adds an admin-gated, host-scoped preview query matching the public homepage data dependencies while bypassing content caches. |
| src/lib/conference/sanity.ts | Threads an opt-in uncached mode through the conference and gallery read paths without changing public defaults. |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
  participant Composer
  participant Preview as Preview iframe
  participant API as conference.homepagePreviewData
  participant Sanity
  Composer->>Preview: Load /admin/homepage-preview
  Preview->>API: Request host-scoped preview data
  API->>Sanity: Uncached conference and gallery reads
  Sanity-->>API: Conference content
  API-->>Preview: Conference and ticket state
  Preview-->>Composer: konf-preview-ready
  Composer->>Preview: konf-preview-state
  Preview-->>Composer: hover, select, and size messages
```

<sub>Reviews (1): Last reviewed commit: ["feat(admin): live homepage preview for t..."](https://github.com/cloudnativebergen/website/commit/329e43f7fad1241ee5717d5c665b6f18551b5763) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=49963321)</sub>

**Context used:**


Starefossen:
Superseded by #741, which fully contains this branch — `git merge-base --is-ancestor` confirms every commit here is already in the workspace branch.

Closing rather than resolving conflicts twice: this branch has six conflicts against current `main` (the content-status and placeholder modules landed separately as #717/#718 and have since taken review fixes, and the conference/gallery data layers have moved under the tenancy work), while #741 merges cleanly. Resolving the same conflicts in both, then re-resolving #741 on top, buys nothing.

The preview surface described here — the client-rendered route, the versioned postMessage protocol, per-band error boundaries, the analytics neutralisation verified three ways, and the Design/Live modes — ships in #741. The description above stays as the record of why each of those exists, since the commit messages carry the detail but not the reasoning.

Issue #687, opened 2026-07-28
contract-signing reads NEXTAUTH_URL — setting it would silently break auth on every tenant domain


Starefossen:
`src/lib/contract-signing/self-hosted.ts` reads **`NEXTAUTH_URL`** as its first base-URL fallback.

That is a trap on a multi-tenant platform. `next-auth/lib/env.js`'s `reqWithEnvURL` rewrites **every request's origin** to whatever `NEXTAUTH_URL` is set to. So an operator who sets it for the entirely reasonable purpose of fixing contract-signing URLs would **silently pin every request to one host and break auth on all other tenant domains** — while appearing to fix an unrelated problem.

The failure is silent: OAuth completes, the browser rejects the mis-scoped cookie, and users on every other domain loop back to sign-in with no error.

This is a live conflict, not a hypothetical: the env var is genuinely useful to the signing module and genuinely destructive to auth.

## Fix
Remove `NEXTAUTH_URL` from the signing module's fallback chain. Use the same resolution the rest of the app uses — `conferenceBaseUrl()` for tenant-scoped URLs, or `NEXT_PUBLIC_BASE_URL` for platform-level ones (the fallback chain already accepts the latter).

## Context
PR #686 adds a system-status check (`auth.fixedOrigin`) that reports an **error** whenever `AUTH_URL`/`NEXTAUTH_URL` is set, with remediation naming `NEXT_PUBLIC_BASE_URL`. That surfaces the hazard but does not remove the reason someone would set it — this issue does.

## Acceptance
- No module reads `NEXTAUTH_URL` or `AUTH_URL`.
- Contract-signing URLs resolve correctly per tenant without any fixed-origin env var.
- A test pins that the signing base URL follows the request host.

Issue #267, opened 2025-11-04
Admin page for managing and merging duplicate speaker profiles across login providers


Starefossen:
**Description**
Implement an admin page to manage duplicate speaker profiles. When speakers log in with different providers (LinkedIn, GitHub), their separate profiles should be merged to ensure data consistency and avoid duplicates.

**Motivation**
This will streamline speaker management, ensuring each speaker has a single profile even when logging in via multiple providers. It solves recurring issues with duplicate speaker entries resulting from multi-provider logins.

**Additional context**
- Admins should be able to search for, review, and merge speaker profiles.
- Automated or semi-automated detection of possible duplicates (e.g. matching name and email).
- Profile merge options for keeping the most recent or most complete data for each field.
- Integrate with current authentication systems (LinkedIn, GitHub).

Starefossen:
**This caused a live incident on 2026.cloudnativedays.no today, and a production census shows it is not isolated.**

## The incident

A speaker reported a confirmed talk missing from their profile, and "speaker not found" when sharing the "I am speaking" banner — while seeing themselves in the speakers list.

Root cause: **two speaker documents for the same person, sharing one slug.**

| Doc | Email | Provider | Created | On the confirmed talk? |
|---|---|---|---|---|
| `1e80d498-4878-4341-9352-00142ce180ec` | ganesh.vasudev@gmail.com | `linkedin:2mtSWuh1kA` | 2026-05-05 | **yes** |
| `241e8419-208a-48c2-8e1b-7080597752d8` | ganesh.vasudevan@ericsson.com | `github:23187057` | 2026-06-15 | no |

Both carry `slug.current == "ganesh-vasudevan"`. He signs in with **GitHub** (the second document), so his dashboard — which filters `$speakerId in speakers[]._ref` — correctly finds no confirmed talk, and the share sidebar (gated on `confirmedTalks.length > 0`) never renders. His *public* profile works, because `getPublicSpeaker` filters on "has a confirmed talk" and so happens to select the LinkedIn document. The person is simultaneously visible and invisible depending on which surface asks.

## Census: 348 speakers, 10 duplicated slugs

`valentin-david` ×3 · `ganesh-vasudevan` ×2 · `eirik-sander-fjeld` ×2 · `marcus-noble` ×2 · `andr-s-valero` ×2 · `kristoffer-dalby` ×2 · `david-pech` ×2 · `cortney-nickerson` ×2 · `graziano-casto` ×2 · `test-user` ×4

[...]

PR #740, opened 2026-08-04
feat(auth): email magic-link sign-in

Starefossen:
Email sign-in, so a conference's speakers and organizers can log in without a GitHub or LinkedIn account. Today those two OAuth providers are the only way in — a real barrier for a general conference audience, and a blocker for the first external customer.

Implements the owner's tiered model: **stateless short-lived tokens for speakers and attendees, stored single-use tokens for organizers and admins**, with the tier chosen from the address at request time.

## It does NOT use an Auth.js adapter — and that was the right call

The plan called for a "minimal five-method adapter". Verified against the installed `@auth/core@0.41.3` source, that is not viable, for two independent reasons:

**1. Any adapter's presence rewrites the existing OAuth flow.** `lib/actions/callback/index.js` calls `adapter.getUserByAccount(...)` **unconditionally** when an adapter is configured, and `handle-login.js` only short-circuits when there is no adapter. So every GitHub and LinkedIn sign-in would start routing through Auth.js's own account-linking, which **throws `OAuthAccountNotLinked`** when a user with that email already exists under another provider — precisely the case this repo deliberately handles by verified-email match in `getOrCreateSpeaker`. That breaks returning users and makes #267 worse.

Also worth recording: the "five methods" are the *asserted* set only. The email path additionally calls `getUser`, `updateUser` and `createUser` at runtime, and the OAuth path calls `getUserByAccount` and `linkAccount`.

**2. The stateless tier is unimplementable on that contract.** `useVerificationToken` receives `{identifier, token: sha256(raw + secret)}` and no request context — a self-describing token cannot be verified from its own hash. The only workaround is smuggling the raw token through AsyncLocalStorage, which is a worse contract violation than not using the contract.

**Implemented instead:** a `Credentials` provider. Credentials + JWT requires no adapter, so `session: {strategy: 'jwt'}`, no adapter, and the entire OAuth path are **byte-for-byte unchanged**.

## The tiers

`resolveEmailLinkTier(email)` returns `stored` if any speaker matching the address — display email **or** `knownEmails` — sits in any conference's `organizers[]`; otherwise `stateless`. It **fails safe to `stored`** on a read error, and is **re-derived at redemption**, so a stateless token minted before a promotion is refused afterwards. The token prefix selects a *verification path*, never a privilege — authorization still comes from `organizerOrgIds` when the session is minted.

- **stateless** — HMAC-SHA256 over `{email, audience host, exp, nonce}` with a domain-separated label, 5 minutes, zero writes.
- **stored** — 32 random bytes, only `sha256(raw + AUTH_SECRET)` persisted, 15 minutes (longer *because* single-use). Consume is a compare-and-swap: uncached read of an unconsumed doc, then a revision-conditioned patch — so exactly one of N concurrent redeemers wins.

## Origin safety

The audience is the **request host**, the same precedence that decides the session cookie's scope — not `conferenceBaseUrl()`, because a tenant is reachable on several hosts at once and that would email a link to a host the user is not on. Carried in the token and compared at redemption.

**Nothing here reads `NEXTAUTH_URL`/`AUTH_URL`, so #687 is not a blocker.** Note next-auth's own `createActionURL` does read them, so a mis-pinned deployment fails **closed** (audience mismatch) rather than authenticating cross-tenant.

## Enumeration, rate limits, account matching

Responses are uniform whether or not the address exists. Rate limits are Sanity-backed with a salted-hash id, so no address or IP is stored at rest: per address 1/60s, 3/15min, 10/24h; per IP 20/h, 60/24h. The address bucket is charged first, so a denial there does not charge the shared IP bucket — NAT protection.

Existing accounts are matched by the same primitives OAuth uses. `getOrCreateSpeaker` was deliberately **not** reused: `computeVerifiedEmails`'s default branch verifies nothing, so it would mint a duplicate on every magic-link sign-in — the #267 failure mode this is meant to avoid.

## Known limitations, stated rather than buried

1. **The email session cookie is host-only** where OAuth widens to eTLD+1, because `signIn()` writes through Next's cookie jar which the per-response `Domain` rewriter never sees. Strictly narrower, never broader — the documented safe degradation — but a magic-link session does not follow the user across sibling subdomains until their next OAuth sign-in self-heals it.
2. **Sender policy genuinely blocks tenant #2.** `resolveEmailSender` had **zero production call sites** before this; on the shared Resend tier a tenant-domain `From` is unverified and the send is rejected. So tenant #2 gets no sign-in mail until it has its own credentials or a verified domain. It now fails **loudly** instead of silently.
3. **Stateless tokens are replayable for their 5-minute life and cannot be revoked** — inherent to the tier decision, and made explicit in the email copy: only stored-tier mail claims single use.
4. A small timing difference between tiers (an organizer costs one extra write), dominated by the Resend call and only exploitable by someone who already knows the address.

5719 tests across 405 files (47 new), covering expired tokens, reuse, a genuine double-redeem race, cross-tenant origin for both tiers, tier confusion, every rate-limit rule, response uniformity and header hygiene. The redemption route was added to the protected-routes smoke script — it is the only surface exercising server-side `signIn` in a real bundle, which is the gap that caused the #462 incident.

<h3>Greptile Summary</h3>

Adds credentials-based email magic-link authentication with stateless speaker tokens and stored, single-use organizer tokens.

- Adds request, delivery, origin binding, rate limiting, redemption, identity matching, and JWT session integration.
- Adds Sanity schemas and cleanup for stored tokens and rate-limit buckets.
- Extends the sign-in UI, email template, documentation, smoke coverage, and focused authentication tests.

The PR should not merge until forced-login redemption and non-atomic stored-token replacement are fixed.

A valid attacker-owned link can establish the attacker’s session in another browser without login-intent binding, while concurrent organizer link requests can leave multiple independently redeemable stored credentials.

**Files Needing Attention:** src/app/api/auth/email-link/callback/route.ts and src/lib/auth/email-link/store.ts

Two security issues were identified: bearer-link GET redemption permits forced login into an attacker-controlled account, and concurrent stored-tier requests can leave multiple organizer credentials active despite the one-live-link guarantee.
</details>

| Filename | Overview |
|----------|----------|
| src/app/api/auth/email-link/callback/route.ts | Bridges bearer-link GET redemption into server-side credentials sign-in, but permits forced login because redemption is not bound to the initiating browser. |
| src/lib/auth/email-link/store.ts | Implements revision-safe per-document consumption, but non-atomic random-ID replacement can leave multiple stored-tier credentials live. |
| src/lib/auth/email-link/rateLimit.ts | Adds privacy-preserving Sanity-backed rolling limits; its documented concurrency overshoot also makes the stored-token replacement race reachable. |
| src/lib/auth/email-link/token.ts | Implements domain-separated HMAC stateless tokens, opaque stored tokens, expiry checks, and host audience binding. |
| src/lib/auth/email-link/verify.ts | Selects the verification path, rechecks organizer tier for stateless links, and consumes stored links before accepting them. |
| src/lib/auth.ts | Adds the credentials provider and projects verified email-link speakers into JWT sessions while preserving the OAuth branches. |
| src/lib/speaker/sanity.ts | Adds verified-email speaker matching and creation; the sole production caller supplies an identifier extracted from a verified token. |
| src/app/(main)/signin/page.tsx | Adds the email sign-in form and generic invalid-link messaging. |
| sanity/schema.ts | Registers the new token and rate-limit document types and exports their Studio-hidden type list. |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
  participant Browser
  participant Action as Sign-in action
  participant Sanity
  participant Mail
  participant Callback as Email-link callback
  participant Auth as Auth.js credentials flow

  Browser->>Action: Submit email
  Action->>Sanity: Rate-limit and resolve tier
  alt Stateless speaker token
    Action->>Action: Sign email, host, expiry, nonce
  else Stored organizer token
    Action->>Sanity: Delete prior token(s), create token hash
  end
  Action->>Mail: Send host-bound link
  Browser->>Callback: GET link with bearer token
  Callback->>Auth: signIn(email-link, token)
  Auth->>Sanity: Verify/consume and resolve speaker
  Auth-->>Browser: Set JWT session cookie and redirect
```

<sub>Reviews (1): Last reviewed commit: ["feat(auth): email (magic-link) sign-in w..."](https://github.com/cloudnativebergen/website/commit/55121be80079ee1f4ff7559a27423a55a9c0be2f) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=49969872)</sub>

> Greptile also left **2 inline comments** on this PR.

<details><summary><h4>Context used (3)</h4></summary>

- Knowledge Base — [Authentication and Access Control](https://app.greptile.com/cloudnativedays/-/custom-context/knowledge-base/cloudnativebergen/website/-/docs/auth-access.md)
</details>


Starefossen:
Security review findings closed in `fb8e2a88`. One of them requires **correcting a claim in the PR description above**, which was materially wrong.

## F1 — the rate limiter, and the wrong disclosure

The PR said the limiter's overshoot was "bounded by a small multiple of the cap". That is false, and the review proved it by running the real limiter against a store modelling Sanity's actual contract:

| | before | after |
|---|---|---|
| burst of 50, cap 1 | **50 allowed**, bucket afterwards recorded **1** hit | 1 allowed |
| 5×20 over 5 minutes, 24h cap of 10 | **60 allowed** | 3 allowed |

`createOrReplace` is a whole-array replace, so concurrency did not overshoot — it **erased the history**. N concurrent requests each read `hits: []` and each wrote `hits: [now]`, so the bucket ended up believing one request had happened, and caps never accumulated across windows. That is a targeted mail-bomb primitive plus a Resend cost and reputation problem.

Now a revision-conditioned compare-and-swap with bounded retry — the same pattern `store.ts` already used correctly for token consumption. Denial happens *before* charging, so a burst cannot keep the window topped up into a permanent lockout. Read failures still fail open (an abuse control must not take sign-in down); **write** failures now fail closed, because a bucket that is never persisted is not a limit at all.

Both scenarios are regression tests.

## F2 — fixed, not just documented

The audience was `x-forwarded-host` with no allowlist, and the review minted a token for `evil.example.net` and replayed it with a spoofed header against a real build — the audience check passed.

The derived host must now be claimed by some conference's `domains[]`, using the routing predicate (exact or single-label wildcard, ports stripped), read live and **failing closed**. Enforced at mint and as **step 0** of verification — deliberately before the stored-tier consume, so a spoofed header can no longer burn a legitimate user's single-use link.

The inherited comment in `origin.ts` is replaced. Its reasoning — that spoofing is harmless because browsers reject mismatched cookie `Domain`s — is sound for cookies and **does not transfer to a bearer token redeemed by curl**.

**Behavioural consequence, documented in `docs/AUTH.md`:** email sign-in now only works on a host a conference actually claims. A brand-new tenant, or a `*.vercel.app` preview, gets no mail until that host is in `domains[]`.

## F3 — login CSRF

Requesting a link now sets an HttpOnly, SameSite=Lax intent cookie holding a salted hash of the token — set on **every** outcome, including a random value when nothing was minted, so its presence is not an enumeration oracle. The GET only calls `signIn` when the cookie matches; anything else (cross-device, or an induced navigation) lands on a `/signin/confirm` interstitial that mints nothing, continues only via an Origin-checked server action, and **peeks** the token without consuming it — so navigating a victim there cannot burn their real link.

Rejected in tests: an intent cookie for a different token, a guessed 64-hex value, and the unsalted `sha256(token)` an attacker holding the token could compute.

## F5 — NFKC divergence

The address proven by delivery and the one written to `knownEmails` are now byte-identical; a divergent input is refused rather than normalised into the verified match-set.

## Not fixed, deliberately

**F4 is pre-existing and filed as #742** — any tenant organizer can rewrite any speaker's display email, which is a login match key. It belongs with the tenancy guards, not here.

**F7 timing oracle is live again.** The review noted it as moot; it is not, now that rate-limit denial exists — a denied request skips the tier lookup, the store write and the Resend call. LOW, and it requires already knowing the address.

Also worth flagging: the branch head was **not prettier-clean** and `format:check` would have failed CI. That churn is in this commit deliberately.

425 files / 5947 tests, production build clean, and `smoke-protected-routes` now 14/14 — up from 11, with the two redemption branches and the interstitial added.

Starefossen:
Both store findings were the same defect and are fixed; the callback finding is already closed on this branch.

**Non-atomic token replacement (greptile + copilot, independently) — real, fixed.** "One live link per address" was enforced by delete-by-query followed by `create` with a random `_id`, which cannot carry it: two overlapping requests both delete, then both create, and each resulting document is independently redeemable by its own hash. That breaks the UX property (a resend should invalidate the previous link) *and* the abuse cap (one document per address no matter how many requests pass the rate limiter).

Now `createOrReplace` on a **deterministic per-address id**, so the later write overwrites the earlier one whatever the interleaving. The delete stays to sweep documents written with random ids before this change — copilot's suggestion exactly.

The id is salted with the same secret as the token hash rather than derived from the address in the clear. Otherwise a document id would be an offline oracle for whether a given person has an account here.

Two regression tests: the same address twice writes one id, and two addresses write different ids.

**Unbound GET forces account login (greptile) — already closed on this branch, before this review ran.** The callback no longer hands a bearer token to `signIn` unconditionally. Requesting a link sets an HttpOnly, SameSite=Lax intent cookie holding a salted hash of the token; `signIn` runs **only** when the presented token matches that cookie. Every other redemption — cross-device, and every induced navigation — is redirected to `/signin/confirm`, which mints nothing, peeks the token without consuming it (so navigating a victim there cannot burn their real link), and continues only through an Origin-checked server action.

See `src/lib/auth/email-link/intent.ts` and the guard at `callback/route.ts:86-107`. Covered by tests that reject an intent cookie for a different token, a guessed 64-hex value, and the unsalted `sha256(token)` an attacker holding the token could compute.

5949 tests green.

Starefossen:
**One real finding fixed; the other three describe code that is no longer here.**

### Fixed: the pending cookie could ship without `Secure` (`793c622b`)

The flag came from `new URL(request.url).protocol === 'https:'`. TLS terminates at the edge, so the handler routinely sees `http://` for a request the browser made over `https://` — meaning the flag would have been **off in production and on in the tests**, which build their requests over `https://`. That cookie carries an unredeemed sign-in token: a bearer credential that must never be allowed onto a plaintext hop.

It now follows `x-forwarded-proto`, the same signal `requestOrigin` already uses to build the emailed link, so the two cannot disagree about whether the tenant is on HTTPS. It falls back to the request URL only when nothing is proxying, so plain-HTTP `next dev` keeps working rather than silently discarding the cookie. Spoofing the header can only turn `Secure` **on**, which fails closed.

Four tests cover it; reverting the fix fails two of them.

### Stale: "unbound GET forces account login"

This describes the login-CSRF hole before the intent-cookie gate. The route now mints a session only when the browser proves it requested the link (`isSameBrowserIntent`); everything else — cross-device, or an attacker-induced navigation — is handed to `/signin/confirm`, whose continue control is a server action. See the `login CSRF on the redemption GET (F3 reproduction)` suite, which asserts `signIn` is *not* called for an unproven navigation, for a mismatched intent cookie, and for a forged one.

### Stale (×2): "token replacement is non-atomic"

Both reviewers describe `delete`-then-`create` with a random `_id`. That is exactly the race the current code was written to remove: it does `createOrReplace` on a **deterministic per-address id** (`storedTokenDocId(identifier)`), so whatever the interleaving, the later write overwrites the earlier and one address can never hold two live links. The `delete` remains only to sweep random-id documents written before the change — and the comment above it says so.

---

Full suite green (5953), lint 0 errors, formatting clean.

PR #731, opened 2026-08-04
fix(security): close the cross-tenant write class across every router

Starefossen:
Closes #730.

The issue named **5** endpoints. Sweeping `src/server/routers/` and `src/app/api/` for the same shape — a document id from client input reaching a patch or delete with no tenant check — found **33 across 10 routers**. `src/app/api/**` is clean: every route-handler write is server-derived or behind a cron-secret / webhook-signature gate a tenant organizer cannot pass.

Not exploitable today (one organization). Live the moment a second customer's data exists. Customers never get direct Sanity access, so these app-level write paths carry the entire load.

## Two found beyond the issue that are worse than the one it named

**`registration.generateToken`** mints a **portal bearer token** from a client-supplied sponsor id — worse than rewriting a document, because it hands over durable cross-tenant read *and* write through the public portal endpoints.

**`speaker.admin.merge`** repoints every inbound reference and then **deletes** the loser, with both ids client-supplied.

## The shared guard

`src/server/tenancy.ts` generalises the gallery-specific helpers from the previous PR: one probe projects every tenancy dimension, and four guards decide — org-level, conference-level, bulk (all-or-nothing), and a speaker variant using the repo's existing membership-or-participation semantics so it admits exactly the set the admin lists already show.

All refuse with an identical `NOT_FOUND`, so a foreign id's existence never leaks through a differing error. `TRPCError` is rethrown through 8 catch blocks that masked the refusal as a 500.

Every guard also constrains `_type` — an ownership check that still permits patching a `conference` document through the *topic* endpoint is wrong even inside one tenant.

Also fixes `/cfp/list`, a still-fail-open speaker-facing read where an unresolvable org returned every tenant's conferences.

## One dataset check before merge

The guards refuse documents carrying no tenant key at all. Safe iff migration 044 ran — which it must have, since admin denies an org-less conference and admin works in production — but it is the single change that could break the present:

```groq
count(*[_type in ["topic", "staff", "sponsor"] && !defined(organization)])
```

## Decisions, not mechanical fixes — flagged for review

- **The sponsor catalog**: `src/lib/sponsor/sanity.ts:317` documents "shared vs partitioned is a pending owner decision" and the *read* pickers tolerate org-less sponsors. The *write* path is now strict.
- **`travelSupport.updateExpenseStatus`** has a deliberate contract change: an unresolvable parent request now refuses rather than proceeding, because the lookup moved from post-write notification to pre-write authorization.
[...]

PR #744, opened 2026-08-04
fix(security): close the cross-tenant leaks hiding in the lint backlog

Starefossen:
Fixes the genuine cross-tenant leaks found while triaging the repo's 254 unread lint warnings. Not cleanup — the annotation work is a separate pass. These are security bugs that had been sitting behind a rule nobody could run.

## The one that matters most: banking PII, and it was reachable

`getAllTravelSupport(conferenceId?)` fell back to `*[_type == "travelSupport"]` — every tenant's speakers' **bank details** — when its argument was falsy.

That was not theoretical. `getConferenceForCurrentDomain()` returns a **truthy `{}`** on an unknown host, so the caller's `if (confError || !conference)` guard never fired and `conference._id` went in as `undefined`. The fallback was one unrecognised Host header away. All four of those guards are now `conference?._id`.

`getSpeakersRequiringTravelSupport` was also rewritten to root at the conference's confirmed talks rather than sweeping every tenant's funding-flagged speakers.

## Cross-tenant bulk writes

`bulkUpdateSponsors` matched `_id in $ids` with no tenant predicate; `bulkDeleteSponsors` issued **no tenant query at all** and passed client ids straight to `transaction.delete`. Both now require a conference, resolve the owned set, and refuse the **whole batch** unless every id resolves — acting on the owned subset would let a crafted batch enumerate another tenant's id space by observing which ids succeeded.

## Another GROQ injection

`getAllWorkshopSignups` was **interpolating client input into GROQ source** with hand-rolled quoting (`_id in ["${id}"]`, `status == "${status}"`) — into the very predicate that carries the tenant boundary. All values are bound now, with a test asserting a hostile id never appears in the query text.

## Global-organizer bridges

`getOrganizers` and `speakerHasStandingInConference` now fail closed, following #728's shape. No `getAllOrganizersAcrossOrgs` escape hatch was added — no caller wants one, and an unused global-read export is a liability.

## Fixed at the data-access layer, deliberately

`src/server/tenancy.ts` does not exist on `main` — #731 is still open — so the shared guard was unavailable. These are scoped in the **library functions** instead, which is strictly stronger: #731 guards the routers, this guards everything underneath, and the two compose as defence in depth. Expect a trivial textual conflict in `src/server/routers/sponsor.ts` if #731 lands first; the semantics agree.

## Two warnings left on purpose

`bulk.ts:273` and the `deleteSponsor` cascade express their tenant predicate as a reference traversal the lint rule cannot parse. Annotating them `groq-global:` would be a lie, so they stay as honest warnings with comments. Warnings in the touched files: 54 → 47.

## Verification

Mutation-checked: deleting each guard fails a named test, and that was actually run. Two honest exceptions are recorded in the test comments rather than claimed as stronger than they are — `getAllTravelSupport`'s guard is backstopped by `scopedFetch`'s empty-scope throw, so only reverting the ternary fails a test.

Single-tenant safety: making these parameters required produced compile errors in **test files only** — every production call site already passed a resolved tenant.

420 files / 5889 tests, typecheck clean, production build passes.

## Action needed before a second tenant's data exists

`sponsorOrgFilter`'s `!defined(organization)` tolerance is **not** fixed, deliberately. Its null-org fail-open half is. But the org-less tolerance is a documented pre-044 bridge, and dropping it would hide every un-backfilled sponsor from the live deployment. Confirm the 044 backfill ran, then delete that clause — it is a genuine leak for org-less rows.

<h3>Greptile Summary</h3>

The PR adds fail-closed tenant scoping to sensitive travel-support, sponsor, speaker, messaging, and workshop operations. Two sponsor bulk-operation ordering gaps remain:
- Mixed bulk-delete batches can cancel owned signing agreements before being refused.
- Closed-won bulk updates can disclose a foreign sponsor's tier state before ownership validation.

The sponsor bulk-operation ordering defects should be fixed before merging because refused requests can cancel valid agreements or disclose foreign sponsor metadata.

Tenant validation occurs only after signing cancellation and after a tier-dependent global preflight, so rejected mixed-tenant batches can still have observable effects.

**Files Needing Attention:** src/server/routers/sponsor.ts

The closed-won bulk-update preflight remains globally keyed by client-supplied ids, allowing response differences to reveal whether a foreign sponsor record has a tier.
</details>

| Filename | Overview |
|----------|----------|
| src/lib/sponsor-crm/bulk.ts | Adds conference-scoped ownership resolution and all-or-nothing checks for bulk sponsor writes. |
| src/server/routers/sponsor.ts | Passes request-derived tenant ids into guarded helpers, but two pre-validation paths still cause side effects or disclose foreign record state. |
| src/lib/travel-support/sanity.ts | Replaces global fallbacks with mandatory conference scoping and roots funding candidates in confirmed talks. |
| src/lib/speaker/sanity.ts | Makes organizer and speaker population queries fail closed when tenant scope cannot be resolved. |
| src/lib/messaging/standing.ts | Removes the global organizer fallback and denies standing for unresolved conferences. |
| src/lib/workshop/sanity.ts | Parameterizes workshop signup filters and applies the conference boundary through scopedFetch. |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
  participant Admin
  participant Router as Sponsor router
  participant Signing as Signing provider
  participant Guard as Tenant-scoped bulk helper
  Admin->>Router: bulkDelete(ownedId, foreignId)
  Router->>Signing: cancel owned agreement
  Router->>Guard: validate batch ownership
  Guard-->>Router: reject mixed batch
  Router-->>Admin: NOT_FOUND
  Note over Signing,Admin: Agreement remains cancelled although deletion was refused
```

<details><summary><h3>Comments Outside Diff (2)</h3></summary>

1. `src/server/routers/sponsor.ts`, line 4553-4557 ([link](https://github.com/cloudnativebergen/website/blob/c6186ec91865e98cb87faf5e040545c5b738baba/src/server/routers/sponsor.ts#L4553-L4557)) 

   <a href="#">P1</a> **Cancellation precedes ownership validation**

   When a bulk-delete request includes an owned sponsor with a pending agreement and a foreign or nonexistent ID, the router cancels the owned agreement before `bulkDeleteSponsors` rejects the mixed batch, leaving the agreement cancelled even though no sponsor records are deleted.

   **Knowledge Base Used:**
   - [Speaker and Sponsor Management](https://app.greptile.com/cloudnativedays/-/custom-context/knowledge-base/cloudnativebergen/website/-/docs/speaker-sponsor-management.md)
   - [tRPC API Layer](https://app.greptile.com/cloudnativedays/-/custom-context/knowledge-base/cloudnativebergen/website/-/docs/trpc-api-layer.md)

2. `src/server/routers/sponsor.ts`, line 4487-4491 ([link](https://github.com/cloudnativebergen/website/blob/c6186ec91865e98cb87faf5e040545c5b738baba/src/server/routers/sponsor.ts#L4487-L4491)) 

   <a href="#">P1</a> <a href="#">security</a> **Tier check precedes tenant validation**

   When a closed-won bulk update contains a foreign sponsor ID, the unscoped tier preflight runs before the conference ownership check, so tierless foreign records return `PRECONDITION_FAILED` while tiered or nonexistent records return `NOT_FOUND`, exposing the foreign record's existence and tier state.

   **How this was verified:** The tier query filters only by client-supplied IDs and branches on its result before the conference-scoped helper validates ownership.

   **Knowledge Base Used:**
   - [Speaker and Sponsor Management](https://app.greptile.com/cloudnativedays/-/custom-context/knowledge-base/cloudnativebergen/website/-/docs/speaker-sponsor-management.md)
   - [tRPC API Layer](https://app.greptile.com/cloudnativedays/-/custom-context/knowledge-base/cloudnativebergen/website/-/docs/trpc-api-layer.md)
</details>

<sub>Reviews (1): Last reviewed commit: ["fix(tenancy): close the fail-open reads ..."](https://github.com/cloudnativebergen/website/commit/c6186ec91865e98cb87faf5e040545c5b738baba) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=50006718)</sub>

- Knowledge Base — [Messaging System](https://app.greptile.com/cloudnativedays/-/custom-context/knowledge-base/cloudnativebergen/website/-/docs/messaging-system.md)
- Knowledge Base — [Workshop management and volunteer sign-up](https://app.greptile.com/cloudnativedays/-/custom-context/knowledge-base/cloudnativebergen/website/-/docs/workshop-volunteer.md)
</details>

### DIFF
--- a/__tests__/api/trpc/sponsor-pipeline-guards.test.ts
+++ b/__tests__/api/trpc/sponsor-pipeline-guards.test.ts
@@ -268,9 +268,12 @@ describe('sponsor CRM pipeline tier invariant — all write paths', () => {
         ids: ['a', 'b'],
         status: 'closed-won',
       } as any)
+      // The REQUEST's conference is passed through as the tenant boundary —
+      // never anything derived from `input`.
       expect(bulkUpdateSponsors).toHaveBeenCalledWith(
         expect.objectContaining({ ids: ['a', 'b'], status: 'closed-won' }),
         mockOrganizer._id,
+        mockConference._id,
       )
     })
   })
--- a/__tests__/api/trpc/sponsor-tenancy.test.ts
+++ b/__tests__/api/trpc/sponsor-tenancy.test.ts
@@ -1,0 +1,138 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest'
+import { appRouter } from '@/server/_app'
+import { getConferenceForCurrentDomain } from '@/lib/conference/sanity'
+import {
+  bulkUpdateSponsors,
+  bulkDeleteSponsors,
+  BulkTenancyError,
+} from '@/lib/sponsor-crm/bulk'
+import { deleteSponsor, deleteSponsorTier } from '@/lib/sponsor/sanity'
+import { clientWrite } from '@/lib/sanity/client'
+
+/**
+ * ROUTER-LEVEL TENANCY REGRESSIONS for the sponsor router (#616, batch A1/A2).
+ *
+ * Two things are pinned here that a library unit test cannot see:
+ *
+ *  1. The REQUEST's tenant reaches the write helpers. The conference/org comes
+ *     from the resolved domain, never from `input` — so a crafted payload
+ *     cannot redirect a bulk write at another tenant.
+ *  2. A fail-closed REFUSAL survives the router's catch blocks as a client
+ *     error (NOT_FOUND), not as a 500. A refusal masked as INTERNAL_SERVER_ERROR
+ *     is indistinguishable from a server fault and hides the guard working.
+ */
+vi.mock('@/lib/conference/sanity')
+vi.mock('@/lib/speaker/sanity')
+vi.mock('@/lib/sponsor-crm/sanity')
+vi.mock('@/lib/sponsor-crm/activity')
+vi.mock('@/lib/sponsor-crm/bulk')
+vi.mock('@/lib/sponsor/sanity')
+vi.mock('@/lib/auth', () => ({ getAuthSession: vi.fn() }))
+vi.mock('next/cache', () => ({ revalidateTag: vi.fn() }))
+vi.mock('@/lib/sanity/client', () => ({
+  clientWrite: { fetch: vi.fn(), patch: vi.fn(), transaction: vi.fn() },
+  clientReadUncached: { fetch: vi.fn() },
+  clientRead: { fetch: vi.fn() },
+}))
+
+const ORG = 'org-test'
+const mockOrganizer = {
+  _id: 'spk-org-1',
+  name: 'Org',
+  email: 'org@test.com',
+  isOrganizer: true,
+  organizerOrgIds: [ORG],
+}
+const mockConference = {
+  _id: 'conf-1',
+  title: 'Test Conf',
+  organization: { _type: 'reference', _ref: ORG },
+}
+
+const createCaller = () =>
+  appRouter.createCaller({
+    session: { user: { email: mockOrganizer.email }, speaker: mockOrganizer },
+    speaker: mockOrganizer,
+    user: { email: mockOrganizer.email },
+  } as any)
+
+beforeEach(() => {
+  vi.clearAllMocks()
+  vi.mocked(getConferenceForCurrentDomain).mockResolvedValue({
+    conference: mockConference as any,
+    domain: 'test.com',
+    error: null,
+  })
+})
+
+describe('sponsor router — the request tenant reaches every bulk write', () => {
+  it('bulkDelete passes the REQUEST conference, not anything from input', async () => {
+    vi.mocked(bulkDeleteSponsors).mockResolvedValue({
+      success: true,
+      deletedCount: 1,
+      totalCount: 1,
+    })
+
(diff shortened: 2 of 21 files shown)


Starefossen:
Both findings addressed.

**`batchConfirmSignups` silently dropped ids — real, fixed.** It counted `succeeded`/`failed` over the signups that *resolved*, but reported `total` as the number of ids the caller sent. Five ids of which two resolved returned `succeeded: 2, failed: 0, total: 5`, and the organizer had no signal that three were dropped.

It now refuses the whole batch when any id fails to resolve. That is the same all-or-nothing rule the sponsor bulk operations in this branch already apply — and for the same second reason, which the reviewer did not raise but which matters more: acting on the resolved subset lets a crafted batch **enumerate another tenant's id space** by observing which ids take effect. `TRPCError` is rethrown so the refusal is not masked as a 500.

**The "admin waist" comment — reworded.** Not a typo: "waist" is house jargon here for the narrow point every request passes through, used in three other files. But the reviewer is right that it is confusing in a security note a reader has to trust, and jargon that needs decoding is a poor choice there. It now says "the admin procedure already gated on".

744 tests green across the server and tRPC suites; typecheck clean.

PR #748, opened 2026-08-04
fix(composer): stop the double scroll, and show real defaults in the inputs

Starefossen:
Two problems the product owner reported after using the composer.

## Double scroll

The workspace sized its panes with `lg:h-[calc(100vh-16rem)]` inside a scrolling document. That `16rem` was a guess at chrome whose height actually varies by breakpoint, by the "using default layout" banner, and by the mobile pane toggle — so the composer was routinely taller than its space, the document scrolled *behind* two panes that were already scrolling, and the header, Save and mode toggles rode off the top. Below `lg` there was no height constraint at all, so the page was the scroller.

The fix lets a page claim the viewport rather than guessing at chrome: a new `shell-fit` variant keyed on a `data-shell-fit="viewport"` marker, which the dashboard shell answers by becoming an `h-dvh` flex column with `overflow-hidden`. Pure CSS via `:has()` — no prop threading through the server layout, and no JS pass that would flash the scrolling layout on first paint. **Pages without the marker are byte-for-byte unaffected.**

`dvh` rather than `vh` deliberately, for mobile browser chrome.

Measured with Playwright at 1600, 1280 and 393: `documentElement.scrollHeight === clientHeight`, exactly two scrollable elements on desktop and one on mobile, and the `h1` stays put after scrolling every internal scroller to the bottom.

## Placeholders now show what the page actually renders

The config inputs said "Your conference name" and "A line under the heading" — instructions that tell you nothing about which band you are editing or what happens if you leave it blank.

Each optional copy field now shows the **real fallback the renderer uses**, imported from the same constants the bands render so they cannot drift. Computed fallbacks are shown rendered with this tenant's values ("Meet the speakers at Nordic Platform Days"), not as templates. Fields with no fallback say so — "No heading — the numbers stand on their own", "No message — the counter disappears".

One of these was simply wrong before: the hero's placeholder claimed "Your conference name", but the hero actually falls back to the tenant's tagline and description.

Long copy is excerpted at a word boundary — 44 chars single-line, 80 for textareas, measured at 393px where the box clips without an ellipsis.

## Verification

426 files / 5978 tests, typecheck, knip and format clean, lint 0 errors with no new warnings, production build succeeds. New stories render inside the **real** `AdminLayout`, which is what proves the shell contract rather than asserting it.

Two notes: the mobile header is now permanently pinned and costs ~340px of an 852px phone screen — correct per the request, but a candidate for compacting if it feels heavy in use. And the fix relies on CSS `:has()`; where unsupported it degrades to exactly today's behaviour.

<h3>Greptile Summary</h3>

The PR removes the composer’s guessed viewport height and establishes a shell-to-workspace flex contract so only the rail or preview scrolls. It also replaces instructional placeholders with excerpts of the actual homepage fallback copy and centralizes the venue heading default.
- Adds a `shell-fit` Tailwind variant activated by the composer’s viewport marker.
- Converts the dashboard shell and composer panes to bounded flex layouts with internal scrolling.
- Derives configuration placeholders from shared homepage defaults and tenant copy.
- Adds focused placeholder tests and real-shell Storybook scenarios.

The PR appears safe to merge with no concrete changed-code failure identified.

The viewport marker activates a complete bounded flex chain whose panes retain dedicated scrolling, while the new placeholders were verified against the current renderer fallback and blank-value behavior.

| Filename | Overview |
|----------|----------|
| src/styles/tailwind.css | Adds a descendant-aware `shell-fit` variant used to activate viewport-constrained shell styles. |
| src/components/common/DashboardLayout.tsx | Establishes the complete viewport-height flex and overflow chain only when a descendant requests shell fitting. |
| src/components/admin/composer/HomepageComposer.tsx | Replaces guessed pane heights with bounded flex sizing and makes each visible pane its own scroll container. |
| src/components/admin/composer/PreviewPane.tsx | Keeps preview controls fixed while allowing the preview body to shrink and scroll within available space. |
| src/components/admin/composer/placeholderCopy.ts | Builds readable placeholder excerpts from the same defaults and tenant values used by homepage renderers. |
| src/components/admin/composer/SectionConfig.tsx | Wires the resolved fallback copy into each optional configuration field. |
| src/lib/homepage/sections.ts | Adds a shared venue heading constant used by both rendering and composer placeholder logic. |
| src/components/homepage/VenueBlock.tsx | Replaces the inline venue fallback with the new shared constant. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
  Shell[DashboardLayout shell] -->|detects data-shell-fit marker| Viewport[Viewport-height flex column]
  Viewport --> Header[Dashboard navigation]
  Viewport --> Main[Flexible bounded main area]
  Main --> Composer[Homepage composer]
  Composer --> Chrome[Fixed composer controls]
  Composer --> Panes[Flexible pane area]
  Panes --> Rail[Composer rail internal scroll]
  Panes --> Preview[Preview internal scroll]
```

<sub>Reviews (1): Last reviewed commit: ["fix(composer): one viewport, one scrollb..."](https://github.com/cloudnativebergen/website/commit/d6a99225cead7f5fe5830a1e0153a2ecde1e6115) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=50032317)</sub>

### DIFF
--- a/src/components/admin/composer/ComposerRail.tsx
+++ b/src/components/admin/composer/ComposerRail.tsx
@@ -32,12 +32,18 @@ import {
 } from '@/lib/homepage'
 import type { SectionContentStatus } from '@/lib/homepage/contentStatus'
 import { SECTION_LABELS, type EditorRow } from '@/lib/homepage/editor'
+import type { PlaceholderConference } from './placeholderCopy'
 import { ComposerSectionCard } from './SectionCard'
 
 export interface ComposerRailProps {
   rows: EditorRow[]
   /** `_key`s whose config panel is open. */
   expanded: ReadonlySet<string>
+  /**
+   * The tenant's title/tagline/description — what the config panels quote as
+   * the copy each band renders when its field is left blank.
+   */
+  conference?: PlaceholderConference
   focusKey: string | null
   hoverKey: string | null
   /** `_key` → what the live site does with that section. */
@@ -77,6 +83,7 @@ export interface ComposerRailProps {
 export function ComposerRail({
   rows,
   expanded,
+  conference,
   focusKey,
   hoverKey,
   statuses,
@@ -163,6 +170,7 @@ export function ComposerRail({
                   index={index}
                   total={rows.length}
                   expanded={expanded.has(row._key)}
+                  conference={conference}
                   focused={focusKey === row._key}
                   hovered={hoverKey === row._key}
                   status={statuses.get(row._key)}
--- a/src/components/admin/composer/HomepageComposer.stories.tsx
+++ b/src/components/admin/composer/HomepageComposer.stories.tsx
@@ -3,6 +3,7 @@ import { http, HttpResponse } from 'msw'
 import { ThemeProvider } from 'next-themes'
 import { screen, userEvent, within } from 'storybook/test'
 import { HomepagePreview } from '@/components/admin/preview'
+import { AdminLayout } from '@/components/admin/AdminLayout'
 import { NotificationProvider } from '@/components/admin/NotificationProvider'
 import { mockFeaturedSpeakers } from '@/components/featuredSpeakers.mocks'
 import type { Conference } from '@/lib/conference/types'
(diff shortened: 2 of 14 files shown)


Starefossen:
Correct — the schema defines `description` as `type: 'text'`, a plain string, and no schema in this repo makes it Portable Text. The comment was simply wrong and is fixed.

The `typeof` guard stays, with the reason restated accurately: this reads dataset content, and the Hero itself renders the value only when it is a string — so a wrong-typed row must show "nothing renders" here rather than a placeholder the page would never produce. That is a defensive read, not a claim about the schema.

PR #737, opened 2026-08-04
fix(tenancy): neutralise the Cloud Native defaults a new tenant inherits

Starefossen:
The platform is about to onboard its first external conference. Today a brand-new tenant's public site still says Cloud Native Days things — in copy, in defaults, in legal text, in analytics, and in signed credentials. This neutralises them.

## ⚠️ DEPLOY GATE — read before merging

**Migration 046 has now been RUN in production** (workflow run 30882197078, 3/3 targets, 12 fields set), so the three live editions carry their theme, background pattern, logos and prospectus copy as stored data. That satisfies the first half of this branch's prerequisite — the `DEFAULT_BACKGROUND_PATTERN` flip to `'none'` and the eight neutral prospectus fallbacks are now safe.

**Migration 047 (new, in this branch) has NOT been run, and this branch must not deploy before it does.** Without it, on all three live sites:

| Change | What breaks |
|---|---|
| Pirsch moved out of the root layout | **Analytics stops silently.** Nothing errors; data just stops arriving |
| Three `/info` FAQ answers become optional fields | Travel directions, speaker dinner and local recommendations disappear |
| `#cndb2025` removed from the social wall | Bergen 2025's venue screen loses its hashtag source |

047 imports its targeting from 046 so the two can never disagree about which document is which edition. It substitutes each document's **own stored `city`** into the templates that interpolated it, and skips with a logged follow-up rather than writing "the city center of undefined". `migrations/046/plan.test.ts` now asserts the tripwire has fired, so CI states the prerequisite rather than leaving it in a PR body.

## The five known items

1. **Bergen geography on `/info`** — the worst of them, because it wasn't obviously wrong: the text was interpolated with whatever city a tenant configured, producing fluent, confident sentences that were simply false about their venue. Now optional fields; absent means the question is omitted. Also fixed a literal `undefined` rendering in the location line.
2. **"the platforms Norway runs on"** — neutral fallback, seeding value deleted.
3. **CNCF background as the platform default** — now `'none'`.
4. **Analytics** — a tenant-editable `analyticsPirschCode`, write-path validated because it is interpolated into a script attribute, and **the fallback is no script at all**. Previously every tenant's traffic reported to one hardcoded property and they received none of their own. The privacy page now discloses Pirsch only when a code exists.
[...]

PR #769, opened 2026-08-04
fix(security): gate the href on two ungated PortableText renders

Starefossen:
**The third instance of this class today**, after the email-template href injection (#727) and the `/info` stored XSS (#737). Found while verifying the `@portabletext/react` 6→7 bump (#768), which is itself clean.

## The hole

Two call sites rendered author-supplied Portable Text with **no `components` map**, so they fell back to the library's default link mark — which emits `value.href` verbatim:

- `src/components/SpeakerProfilePreview.tsx:223` — a talk description
- `src/components/proposal/ProposalDetailsForm.tsx:134` — a proposal abstract

Both render content the speaker wrote. Every other `<PortableText>` in the repo passes `portableTextComponents`, whose `link` mark runs each href through `toSafeRichTextHref`.

**React is not a substitute for that gate.** It scrubs a bare `javascript:` href, but `data:`, `vbscript:`, protocol-relative `//evil.example` and `https:evil.example` pass straight through — verified by sabotaging the gate and observing what survived.

## Why nothing caught it

The omission is invisible to type-checking — `components` is optional — and invisible to any rendering test that does not happen to feed it a hostile href. It was found by reading call sites, which is not a control.

So the fix is not just the two props: **`__tests__/lib/portabletext/gated-renders.test.ts` scans every `.tsx` under `src/` and fails if any `<PortableText>` omits `components`.**

Two things make that scan trustworthy rather than decorative:
- It tracks **brace depth** when finding each element's closing bracket, so an attribute containing `>` (an arrow function, e.g. `({ children }) => …`) cannot end the match early and silently skip elements.
- It asserts it found **at least 8** renders, so a parser that stopped matching would fail loudly instead of passing on an empty set.

## Verification

| state | result |
| --- | --- |
| both sites gated | 2/2 pass |
| one site reverted to the ungated form | **1 failed**, naming `ProposalDetailsForm.tsx` |
| restored | 2/2 pass |

Full suite **6534 passed** · `tsc --noEmit` clean · `eslint .` 0 errors / 216 warnings · `prettier --check` clean.

## Scope

No behaviour change for well-formed content — the gate only rewrites hrefs it refuses. Ungated rendering was the bug.

<h3>Greptile Summary</h3>

This PR adds the shared Portable Text components map to two previously ungated author-content renders and introduces a source scan intended to prevent future omissions. The two production fixes use the canonical href gate, but the scan accepts any components map and therefore misses an existing inline raw-href renderer.

- Gates talk descriptions in `SpeakerProfilePreview`.
- Gates read-only proposal abstracts in `ProposalDetailsForm`.
- Adds a repository-wide Portable Text render scan.

The two targeted render fixes are sound, but the repository-wide security control must be corrected because it currently passes while the same unsafe-link behavior remains reachable in the homepage hero.

The scan treats every components map as gated even though an existing inline link renderer writes author-controlled hrefs directly, undermining the security guarantee introduced by this PR.

**Files Needing Attention:** __tests__/lib/portabletext/gated-renders.test.ts and src/components/Hero.tsx

The new regression test does not establish that a render uses the safe link handler: `Hero.tsx` supplies an inline components map but emits `value.href` directly, so hostile organizer-authored links remain reachable while the test passes.

**How this was verified:** The scan checks only for `components=`, while the Hero link mark assigns `value?.href` directly instead of calling `toSafeRichTextHref`.
</details>

| Filename | Overview |
|----------|----------|
| __tests__/lib/portabletext/gated-renders.test.ts | Adds a non-vacuous source scan, but checks only for a components prop and false-passes inline maps that do not gate links. |
| src/components/SpeakerProfilePreview.tsx | Correctly applies the canonical gated components map to Portable Text talk descriptions. |
| src/components/proposal/ProposalDetailsForm.tsx | Correctly applies the canonical gated components map to read-only proposal abstracts. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart LR
  A["Organizer-authored announcement"] --> B["Hero PortableText"]
  B --> C["Inline components map"]
  C --> D["link mark: href=value.href"]
  D --> E["Ungated anchor"]
  T["New source scan"] --> P{"components= present?"}
  P -->|Yes| G["Test passes"]
```

<sub>Reviews (1): Last reviewed commit: ["fix(security): gate the href on two unga..."](https://github.com/cloudnativebergen/website/commit/6a320a620e4c5a9b9fa3b0f0948453eb0e0afd1e) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=50114474)</sub>

**Context used:**

- Knowledge Base — [Call for Papers: Proposal Submission and Review Flow](https://app.greptile.com/cloudnativedays/-/custom-context/knowledge-base/cloudnativebergen/website/-/docs/proposal-cfp.md)

### DIFF
--- a/__tests__/lib/portabletext/gated-renders.test.ts
+++ b/__tests__/lib/portabletext/gated-renders.test.ts
@@ -34,6 +34,26 @@ function tsxFiles(dir: string, out: string[] = []): string[] {
 }
 
 /**
+ * Blank out comments, preserving offsets and line structure.
+ *
+ * Without this the scan matches prose. A doc comment that mentions
+ * `` `<PortableText>` `` looks exactly like a render whose very next character
+ * is `>`, so the walk below captures an empty element, finds no `components`
+ * attribute, and reports a false positive against a file that is in fact
+ * correct — which is what happened the first time this ran in CI.
+ *
+ * Replacing rather than deleting keeps every other index valid.
+ */
+function stripComments(source: string): string {
+  return source
+    .replace(/\/\*[\s\S]*?\*\//g, (m) => m.replace(/[^\n]/g, ' '))
+    .replace(
+      /(^|[^:])\/\/[^\n]*/g,
+      (m, lead) => lead + ' '.repeat(m.length - lead.length),
+    )
+}
+
+/**
  * The attribute text of each `<PortableText …>` element in `source`.
  *
  * Scans to the element's own closing bracket while tracking brace depth, so an
@@ -41,7 +61,8 @@ function tsxFiles(dir: string, out: string[] = []): string[] {
  * early — a regex stopping at the first `>` would silently skip elements and
  * make this whole test pass vacuously.
  */
-function portableTextElements(source: string): string[] {
+function portableTextElements(input: string): string[] {
+  const source = stripComments(input)
   const elements: string[] = []
   // `\b` alone would also match PortableTextEditor / PortableTextBlock.
   const opening = /<PortableText(?![A-Za-z0-9_])/g
@@ -69,7 +90,8 @@ function portableTextElements(source: string): string[] {
  * exactly that, and the presence-only check above accepted it — so this second
  * scan looks at what the mark DOES, not merely that a map was supplied.
  */
-function linkMarkBodies(source: string): string[] {
+function linkMarkBodies(input: string): string[] {
+  const source = stripComments(input)
   const bodies: string[] = []
   const opening = /\blink:\s*(\(|function)/g
   let match: RegExpExecArray | null
--- a/__tests__/lib/portabletext/gated-renders.test.ts
+++ b/__tests__/lib/portabletext/gated-renders.test.ts
@@ -1,0 +1,94 @@
+/**
+ * EVERY `<PortableText>` render must pass the gated `components` map.
+ *
+ * The library's DEFAULT link mark emits `value.href` verbatim. Portable Text
+ * here is author-supplied — speaker bios, talk abstracts, sponsor terms — so an
+ * ungated render turns stored content into a live anchor with an attacker's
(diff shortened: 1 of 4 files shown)


Starefossen:
**The reviewer was right, and it found a fourth ungated site.** Fixed in `e62a4d83`.

Checking that `components=` is *present* proves nothing about what the map does. `src/components/Hero.tsx` passes an **inline** map whose link mark assigned `value?.href` directly — so `conference.announcement`, which is author-supplied, rendered a live anchor with the author's scheme, and my new test accepted it.

Worse than a miss: a presence-only check reads like coverage. It would have gone on reporting a clean sweep while the homepage rendered ungated links.

**Two changes:**

1. The Hero's link mark now routes through `toSafeRichTextHref`, and derives `target`/`rel` from the **safe** href rather than the raw one — otherwise a refused scheme could still open a new browsing context.
2. The scan now asserts every `link:` mark that emits an `href` routes through `toSafeRichTextHref`, so an inline map cannot smuggle a raw one past it.

**A note on the parser, because it nearly repeated the mistake it exists to prevent.** My first version walked to the first balanced bracket, which ends at the *parameter list* — `({ children, value })` — and never reached the body where the href is assigned. It matched **zero** marks. The anti-vacuity assertion (`expect(found).toBeGreaterThanOrEqual(2)`) failed loudly instead of reporting a clean sweep over an empty set. Without it, this PR would have shipped a security scan that silently checked nothing. It now walks to the end of the object property.

**Verification:**

| state | result |
| --- | --- |
| all four sites gated | 3/3 pass |
| Hero reverted to the raw href | **1 failed**, naming `src/components/Hero.tsx` |
| restored | 3/3 pass |

Full suite **6535 passed** · `tsc` clean · `eslint` 0 errors / 216 warnings · `prettier` clean.

PR #771, opened 2026-08-04
feat(schema): lock conference and organization append-only for kontroll

Starefossen:
## Why

A second application is now being built against this Sanity dataset:
**`RunKonf/kontroll`**, the self-service control panel at my.konf.app. It reads
`conference` to list an organization's editions and reads/writes `organization`.

It does **not** compile against this repository. So when a field here is deleted
or retyped, nothing breaks at build time — it breaks that app **silently, at
runtime, in production, against live content**. `conference` has ~160 top-level
fields and grows most weeks. This PR makes that drift impossible to introduce by
accident.

## 1. The narrow cross-app read contract

**`src/lib/conference/contract.ts`** — one place, and obviously the contract.

- `CONFERENCE_LIST_PROJECTION` — the GROQ projection kontroll runs
- `CONFERENCE_LIST_BY_ORGANIZATION_QUERY` — the full query, `$organizationId`
- `ConferenceListItem` — the matching row type
- `CONFERENCE_CONTRACT_FIELDS` — the schema fields it depends on, asserted below

The point is that kontroll depends on **nine schema fields** (plus the system
field `_id`, which the schema does not declare — ten projection entries, nine
schema fields), not 80. The whole module is standalone — no `@/lib` imports, no Sanity client, no Next.js — so it can be
lifted verbatim into a shared `@runkonf/core` package later. A comment in the
file says so.

Two corrections against the requested field list, made after checking the real
schema:

- **`lifecycle` is not a field.** It is a Studio *fieldset* (an editor grouping).
  The field is **`lifecycleStatus`**, and that is what the contract names.
- `visibility` and `domains` are **coalesced in the projection**
  (`absent -> "live"`, `absent -> []`), so a consumer never has to re-derive
  absent-means-live — the rule documented in `src/lib/conference/visibility.ts`.

`organization` is projected as `"organizationId": organization._ref`: the
consumer already knows which org it asked for and needs the key, not the doc.

## 2. The append-only lock

**`__tests__/sanity/schema-contract.test.ts`** snapshots both document types
against **`sanity/schema-shape.baseline.json`**.

The walker flattens a document type to a sorted `path -> type` map. It follows
`fields` and `of` — inline objects, arrays of objects, their nested arrays — and
**resolves named types**: a field whose `type` is a separately registered schema
type (`richTextCode` and the rest of the rich-text vocabulary) carries no inline
fields, so the walker looks it up in `schema.types` and continues underneath it.
Without that, `richTextCode.code` would sit outside the lock.

**Cycle guard:** resolving named types can loop (`blockContent` embeds types that
embed `blockContent`). The walk carries the set of type names already resolved on
the current path and refuses to resolve one twice; the repeat still gets its own
`path -> type` entry, so removing it is still caught — only the descent stops.
The guard is per-path, not global, so the same type under two different fields is
expanded under both. Covered by two unit tests: a mutual `A -> B -> A` cycle and
a self-embedding type.

The comparison is **asymmetric on purpose**: paths in the
current schema that are absent from the baseline are additions and are never
reported. Only removals and type changes are drift. A rename is a removal plus
an addition and fails on the removal.

It is also **order-independent** (a map, not a list) — reordering fields breaks
no reader, so it is not a failure.

On top of the snapshot, every field named in `CONFERENCE_CONTRACT_FIELDS` is
asserted to still exist, with its **own** failure message naming the missing
field. That is the tightest part of the contract.

### The escape hatch

The rule is meant to be overridable — just not accidentally. A rule that cannot
be overridden gets deleted the first time it is inconvenient. The failure message
says, in full:

```
The "conference" schema is APPEND-ONLY. A SECOND application —
RunKonf/kontroll, the control panel at my.konf.app — reads these documents
straight out of Sanity. It does not compile against this repository, so
removing or retyping a field does not break its build: it breaks that app
SILENTLY, AT RUNTIME, in production, against live content. Adding fields is
always fine and this test will not complain about it.

Renaming counts as a removal plus an addition, and fails on the removal.

IF THE REMOVAL IS INTENDED, this is the escape hatch — the rule is meant to
be overridable, just not accidentally:

  1. Make sure kontroll no longer reads the field.
  2. Run:  pnpm tsx scripts/update-schema-baseline.ts
  3. Commit the regenerated sanity/schema-shape.baseline.json IN THE SAME PR,
     so a reviewer sees the removal as an explicit line in the diff.
```

## The baseline is generated, not hand-written

Written only by `scripts/update-schema-baseline.ts`, which imports the real
schema modules:

| document type  | paths captured | of which top-level |
| -------------- | -------------- | ------------------ |
| `conference`   | **235**        | 77                 |
| `organization` | **18**         | 10                 |

Nested capture is real, e.g.
`homepageSections[homepageFaq].items[homepageFaqItem].answer: text` — an array
of objects, inside an array of objects, three levels down — and named-type
capture is real too, e.g.
`homepageSections[homepageRichText].content[richTextCode].code: text`, which
exists only because the walker resolves `richTextCode` out of the registry.

## Sabotage evidence

A test that cannot fail is worthless, so all four directions were sabotaged
against the real schema and restored. Observed results:

| # | Sabotage | Expected | Observed |
| - | -------- | -------- | -------- |
| 1 | Added `sabotageDummyField: string` to `conference` | PASS | **PASS — 20/20** |
| 2 | Removed `tagline` from `conference` | FAIL | **FAIL — 1 failed / 19 passed**, `Fields REMOVED (1): - tagline` |
| 3 | Retyped `startDate` `date -> datetime` | FAIL | **FAIL — 1 failed / 19 passed**, `Fields RETYPED (1): ~ startDate: date -> datetime` |
| 4 | Removed `domains` (a projection field) | FAIL, specific message | **FAIL — 2 failed / 18 passed**: the generic snapshot diff *and* `MISSING CONTRACT FIELD: "domains" is named by the cross-app projection...` |
| 5 | Removed `richTextCode.code` — a nested field of a **separately registered named type** | FAIL | **FAIL — 1 failed / 24 passed**, `Fields REMOVED (1): - homepageSections[homepageRichText].content[richTextCode].code` |
| 6 | Retyped `richTextCode.code` `text -> string` | FAIL | **FAIL — 1 failed / 24 passed**, `~ homepageSections[homepageRichText].content[richTextCode].code: text -> string` |

Plus one bonus, to prove nesting and `organization` are genuinely covered:
removing `featureOverrides[].note` from `organization` failed with
`Fields REMOVED (1): - featureOverrides[featureOverride].note`.

Sabotages 5 and 6 are the review finding: before named-type resolution neither
path existed in the shape at all, so both edits passed the lock silently. Sabotage
1 was re-run after the walker change to confirm append still passes.

Baseline (nothing sabotaged): **25/25 pass**; full suite **6567 passed / 463
files**.

The lock's four directions are also unit-tested against synthetic shapes inside
the same file, so the evidence lives in CI rather than only in this description.

## Known limit

Traversal follows `fields`, `of`, and named types resolved through the registry.
Portable-text internals reached through `marks.annotations` (rich-text link
annotations, for instance) are **not** walked — block-content plumbing, not
fields any cross-app reader projects. This is stated in the walker's and the
test's header comments too, not just here.

One coupling worth knowing: the homepage section `variant` fields are generated
from `SECTION_VARIANTS`, so dropping a section type below two variants removes
`variant` from the shape and trips the lock. That is arguably correct — it *is* a
schema shape change — but it will surprise someone one day.

## Gates

`pnpm test` (6567 passed / 463 files) · `pnpm typecheck` · `npx eslint .` (0
errors, 216 warnings) · `npx prettier --check .` · `pnpm knip` · `pnpm build`
with the CI placeholder env — all green.

`src/lib/conference/contract.ts` is added to `knip.json`'s ignore list: its
consumer is another repository, so knip correctly sees its exports as unused.


Starefossen:
**Both findings fixed and independently verified.**

**Named types now nest.** I reproduced the reviewer's gap and confirmed the fix rather than taking it on trust: removing the `code` field from the separately-defined `richTextCode` type now fails the lock, naming the exact path —

```
Fields REMOVED (1):
  - homepageSections[homepageRichText].content[richTextCode].code
```

25/25 with the field present, **1 failed / 24 passed** with it removed, 25/25 restored. Before the fix that sabotage passed silently, which is what made the header's "full depth" claim an overstated guarantee rather than a documentation nit — the sort that gets trusted precisely when it matters.

**Field count corrected** — nine schema fields plus `_id`, not ten schema fields.

Also verified the baseline genuinely captures the deeper structure now (18 `richText*` paths present, including `homepageSections[homepageRichText].content[richTextCode].{code,filename,language}`), so the coverage is real rather than asserted.

PR #772, opened 2026-08-04
chore(deps): bump sanity, @sanity/types and @sanity/vision to v6

Starefossen:
Bumps the three version-coupled Sanity Studio packages together:

| package | from | to (range) | resolves |
| --- | --- | --- | --- |
| `sanity` | `^5.28.0` | `^6.8.0` | 6.9.0 |
| `@sanity/types` | `^5.28.0` | `^6.8.0` | 6.8.0 |
| `@sanity/vision` | `^5.28.0` | `^6.8.0` | 6.9.0 |

**No source changes were required.** The diff is `package.json` + `pnpm-lock.yaml` only.

## Breaking changes in Studio v6, and how each lands here

Studio v6 (released 2026-06-11) is a deliberately narrow major. Per the [v6 changelog](https://www.sanity.io/docs/changelog/studio-NS4zMS4w.md) and the [Studio v6 post](https://www.sanity.io/blog/sanity-studio-v6):

| breaking change | impact here |
| --- | --- |
| **Node >= 22.12** (Node 20 dropped) | None. Every workflow pins `node-version: '24'`, and `.node-version` is `24`. |
| **`auth.providers` `mode` option removed** (`mode: 'append'` no longer supported) | None. `sanity.config.ts` sets no `auth` config at all. |
| **Default search strategy `groqLegacy` -> `groq2024`** | Studio-only behaviour, and no `search` config is set. Does not touch app GROQ, which goes through `@sanity/client` directly. |
| **React strict mode on by default in dev** | Dev-only Studio behaviour. |
| **Vite 8 for the Studio build** | Studio build got substantially faster (see below). Storybook's Vite also moved 8.0.14 -> 8.2.0 as a dedupe side effect; `pnpm build-storybook` passes. |

Notably **not** changed in v6: `defineType` / `defineField` / `defineArrayMember`, the `Rule` validation API, and the `@sanity/types` type surface (`SanityDocument`, `PortableTextBlock`, `Reference`, `Image`, `TypedObject`, `SchemaTypeDefinition`). That is why `sanity/schemaTypes/` and every app type built on it compile untouched.

## Peer compatibility

No peer bumps needed:

- `next-sanity` 13.3.0 already declares `sanity: ^5.29.0 || ^6.0.0` — stays put.
- `@sanity/client` stays at 7.26.0 (`sanity@6.9.0` depends on `^7.26.0`).
- `@sanity/migrate` stays at `^6.1.2` — **deliberately not bumped**, see below.
- `next` (^16.3.0), `typescript` (^6.0.3) and `knip` (^6.15.0) untouched.
- `@starefossen/sanity-plugin-inline-svg-input` declares `sanity: ^3 || ^4`, so its peer was already unmet at Sanity 5 — unchanged by this PR, and the plugin's `inlineSvg` type still compiles into the schema (proven below).

## ⚠️ Please read: `@sanity/migrate` runtime moves 6 -> 8 as a side effect

The direct devDependency is **not** bumped and stays `@sanity/migrate: ^6.1.2`. But `sanity` itself carries its own copy, and that copy changed major:

- `sanity@5.31.1` depended on `@sanity/migrate: ^6.1.2`
- `sanity@6.9.0` depends on `@sanity/migrate: ^8.0.1`

`run-migration.yml` executes migrations via `pnpm sanity migration run`, i.e. through the CLI — so **the engine that runs production migrations becomes v8 with this bump**, and that is unavoidable while taking Studio v6. Two consequences:

1. The tree now holds two copies: `@sanity/migrate@8.0.1` (what the CLI runs) and `@sanity/migrate@6.1.2` (the direct devDependency, imported only by `migrations/002-proposal-desc-to-portable-text/index.ts` for `at`/`defineMigration`/`set`).
2. That direct 6.1.2 copy's `@sanity/cli-core: ^1` peer is now resolved against `@sanity/cli-core@2.8.0` (pulled up by Sanity 6), so it is an unmet peer. pnpm does not warn, and `pnpm install --frozen-lockfile` is clean.

**Verified the migration path still works**, by running the exact workflow command against both trees:

```
# Sanity 5 (main) and Sanity 6 (this branch) - IDENTICAL behaviour
npx sanity migration run 047-tenant-defaults-backfill --project ... --dataset production
  -> Running migration "047-tenant-defaults-backfill" in dry mode
  -> HTTPError: Unauthorized: Session not found     (expected: fake token)
```

The migration loads and the dry run starts on both; it only stops at the auth boundary. Sanity 6 actually reports the error more cleanly (no raw stack dump).

Separately noted, **pre-existing on `main` and not a regression**: `sanity migration list`, and `sanity migration run 002-proposal-desc-to-portable-text`, both fail identically on Sanity 5 and Sanity 6 with `ParseError: Unexpected token` on `src/lib/proposal/ui/badges.tsx` — migration 002 transitively imports a `.tsx` file the migrate loader cannot parse. Worth its own issue.

If you would rather the migration engine not move in this PR, this PR has to be held — Studio v6 cannot be taken without it.

## Verification

Everything below was run on a real `pnpm install` (not a symlinked `node_modules`); `sanity@6.9.0` confirmed installed.

| check | result |
| --- | --- |
| `pnpm typecheck` | clean, **0 errors** — no `any`, no `@ts-expect-error`, no tsconfig loosening was needed |
| `pnpm test` | **462 files / 6542 tests passed** |
| `npx eslint .` | **0 errors, 216 warnings** (unchanged baseline) |
| `npx prettier --check .` | clean |
| `pnpm knip` | exit 0 (only the pre-existing `@workos-inc/node` config hint) |
| `pnpm build` | exit 0 |
| `pnpm build-storybook` | completed successfully |
| `npx sanity build` (what `deploy-studio.yml` runs) | **built in 759 ms**, 18 MB output — the Vite 8 speedup is real |
| `pnpm install --frozen-lockfile` from scratch | clean, no peer warnings |

`pnpm typecheck` and `pnpm test` were re-run after a clean `--frozen-lockfile` reinstall, so the results above reflect exactly the tree this lockfile produces.

## Proving the schema still validates

`sanity/schemaTypes/` is the data model for every tenant, so "it compiles" is not sufficient — a schema can compile while silently dropping a validation rule. Two executed proofs, not code reading:

**1. `conference` required fields are still enforced.** Built a real workspace from the actual `sanity.config.ts` (plugins included, so the `inlineSvg` type resolves) via `createWorkspaceFromConfig`, then ran `validateDocument()` from `sanity`:

```
compiled conference required fields (18): title, organization, organizer, city,
  country, startDate, endDate, cfpStartDate, cfpEndDate, cfpNotifyDate,
  programDate, contactEmail, cfpEmail, sponsorEmail, formats, topics,
  organizers, domains

EMPTY conference          -> 18 validation errors  (one "Required" per field above)
MISSING-title conference  -> title errors: 1 ["Required"]
WITH title                -> title errors: 0
```

So a `conference` document missing `title` is rejected, and adding `title` clears exactly that error. **The same script produces byte-identical output on `main` (sanity 5.31.1).**

**2. The whole schema's rule surface is unchanged.** Walked every compiled type in the workspace and dumped every validation rule spec (flag + constraint) at every field path, on both trees:

```
sanity 5.31.1 -> TOTAL rule-bearing paths: 888
sanity 6.9.0  -> TOTAL rule-bearing paths: 888
diff rules-v5.txt rules-v6.txt -> IDENTICAL
```

Not one validation rule was added, dropped, or altered by the upgrade.

## Risk notes

- The Studio is **not** embedded in the Next app (no `NextStudio` route); it is standalone and deployed by `deploy-studio.yml`. So Studio v6 carries no app-runtime risk — its blast radius is that workflow, which is verified building.
- `@sanity/types` resolves to 6.8.0 while `sanity` bundles 6.9.0. Multiple `@sanity/types` copies already existed on `main` (5.28.0 / 5.31.1 / 6.3.0), and `tsc --noEmit` is clean, so the skew is benign.

Do not merge or enqueue — flagging the `@sanity/migrate` engine move for a decision first.

<h3>Greptile Summary</h3>

This PR upgrades the version-coupled Sanity Studio packages from v5 to v6 and refreshes their transitive dependency graph.
- Upgrades `sanity`, `@sanity/types`, and `@sanity/vision`.
- Moves the Studio toolchain to its newer CLI, Vite, and migration-runtime dependencies.
- Retains the direct `@sanity/migrate` v6 dependency used by migration source files.

The dependency upgrade appears safe to merge based on the reviewed configuration, workflow constraints, resolved peers, and documented verification.

Node 24 satisfies the upgraded toolchain, frozen-lockfile installs prevent dependency drift in automated environments, and no concrete incompatible Sanity configuration, peer relationship, build path, or migration behavior remains.

| Filename | Overview |
|----------|----------|
| package.json | Upgrades the three version-coupled Sanity packages to compatible v6 ranges without exposing a concrete repository incompatibility. |
| pnpm-lock.yaml | Resolves the Sanity v6 dependency graph, including the acknowledged migration-engine and build-tool updates, with no actionable regression established. |

</details>

<sub>Reviews (1): Last reviewed commit: ["chore(deps): bump sanity, @sanity/types ..."](https://github.com/cloudnativebergen/website/commit/e841b558d528eb3b3a161a0f00195327a06fd9bb) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=50136746)</sub>


Starefossen:
**Confirmed with Hans: merging, with a dry-run gate on the next migration.**

Both this PR's own analysis and the review above independently found the same thing: `sanity` v6 pulls `@sanity/migrate` 8.x transitively via `@sanity/cli`, and `run-migration.yml` executes `pnpm sanity migration run` — so the **production migration engine** moves to v8 whether or not the direct devDependency does. Holding the direct dependency back does not prevent it; a separate `@sanity/migrate` PR cannot either.

**Why merge rather than block:** the change is unavoidable with Studio v6, and the evidence for the rest of the upgrade is strong — 888 rule-bearing validation paths byte-identical between trees, `conference` still compiling exactly 18 required fields with identical errors, `npx sanity build` (what `deploy-studio.yml` needs) building clean, and the migration CLI loading and starting a dry run identically on both trees.

**What that evidence does NOT cover, stated plainly:** the CLI check stops at `Unauthorized` with a fake token. It proves resolution and startup, not that a real migration emits identical patches. No production credentials were available to go further.

**So the gate moves to the next migration run:** dry-run first, compare the reported document/field counts against expectations, and only then apply — the same discipline used for 046 and 047. The workflow already does dry-run → backup → apply; this just makes the comparison deliberate rather than assumed, because it will be the first production migration on a new engine major.

Also worth recording from this PR: `sanity migration list` and `migration run 002-…` fail identically on **both** v5 and v6 with a `ParseError` on `src/lib/proposal/ui/badges.tsx` — migration 002 transitively imports a `.tsx` the loader cannot parse. Pre-existing, not a regression, and worth its own issue.

Issue #752, opened 2026-08-04
Pirsch analytics code: three silent-misconfiguration gaps (edition clone, privacy disclosure, admin hint)


Starefossen:
Three small defects around the `analyticsPirschCode` field introduced in #737. None is a security issue; all three are silent-misconfiguration bugs, which is the class that costs an organizer a season of data before anyone notices.

`resolvePirschCode()` (`src/lib/analytics.ts`) is the single source of truth for whether a stored code is usable — it trims and tests against `PIRSCH_CODE_PATTERN`, returning `undefined` for anything malformed. The runtime script injection uses it. Three other places do not.

### 1. Creating the next edition silently drops the analytics code

`src/lib/conference/edition.ts` builds the new conference document by explicit per-field enumeration (`...(src.venueName ? { venueName: src.venueName } : {})`, around line 336). `analyticsPirschCode` was never added to that list, so an organizer who creates next year's edition gets a site that quietly stops collecting page views until someone re-enters the code by hand.

Analytics config belongs with the identity fields that are always carried (city/country/venue/logos) — it is the same Pirsch site across editions.

Needs a test: the edition-clone tests should assert the field carries, otherwise the next new conference field will land in the same hole.

### 2. The privacy policy can disclose a subprocessor that is never contacted

`src/app/(main)/privacy/page.tsx` (~line 1014) gates the Pirsch subprocessor row on `conference?.analyticsPirschCode` truthiness. With a malformed value stored, the runtime serves no script but the published privacy policy still names Pirsch as a processor of visitor data.

Over-disclosure rather than under-disclosure, so the exposure is legal-accuracy, not privacy — but a privacy policy that describes processing which does not happen is still wrong, and it is the document tenants are most likely to be held to.

### 3. The admin settings hint claims analytics is on when nothing is served

`src/app/(admin)/admin/settings/page.tsx` (~line 633) shows "No analytics script is served…" only when the field is falsy. Store a typo'd code and the UI goes quiet — implying analytics is configured — while the public site serves nothing. The organizer has no signal until they notice an empty dashboard.

### Fix

Gate all three on `resolvePirschCode(...)` rather than truthiness, and add `analyticsPirschCode` to the always-copied identity fields in the edition builder. Small, but worth a test each: the divergence between "a value is stored" and "a value is usable" is what produced all three.

Found by automated review on #737; deliberately not folded into that PR, which was carrying a stored-XSS fix.

PR #775, opened 2026-08-04
feat(provisioning): bearer-authenticated tenant creation API for kontroll

Starefossen:
## What

`RunKonf/kontroll` (my.konf.app) is a separate app with no Konf session, so it cannot call the onboarding wizard's `platformProcedure` mutation. This adds a **machine door to the same transaction** — not a second implementation.

`src/lib/onboarding/provision.ts` now owns `provisionOrganization()`: the global slug/domain uniqueness reads, the single-match organizer resolution, and the all-or-nothing Sanity transaction that were previously inline in `src/server/routers/onboarding.ts`. Both the tRPC mutation and the new route call it. A duplicate would have forked the per-field document enumeration in `buildOnboardingDocuments` and started silently dropping new conference fields on whichever path was forgotten (#752).

## Contract

```http
POST /api/provisioning/organizations
Authorization: Bearer <PROVISIONING_API_TOKEN>
Idempotency-Key: <opaque, 16-200 printable ASCII chars>   # REQUIRED
Content-Type: application/json
```

Body is `CreateOrganizationSchema` verbatim — the **same** Zod schema the wizard posts through, not a looser copy:

```json
{
  "organization": { "name": "...", "slug": "...", "contactEmail": "...", "billingEmail": "?" },
  "conference":   { "title": "...", "city": "...", "country": "...", "startDate": "?", "endDate": "?" },
  "organizer":    { "name": "...", "email": "..." },
  "domains":      ["oslo.example.com"]
}
```

| Status | Body |
| --- | --- |
| `201` | `{ organizationId, conferenceId, speakerId, speakerCreated, replayed: false, challenges: DomainVerificationView[] }` |
| `200` | same shape, `replayed: true` — key already used, original ids returned, nothing written |
| `400` | `{ error: "invalid_request", code, issues? }` — `idempotency_key_required` / `invalid_json` / `schema_validation_failed` |
| `401` | `{ error: "unauthorized" }` — **uniform for every auth failure** |
| `409` | `{ error: "conflict", code, slug? \| domains? }` — `slug_taken` / `domain_claimed` / `ambiguous_organizer` |
| `429` | `{ error: "rate_limited" }` + `Retry-After` |
| `500` | `{ error: "internal_error" }` |

**Env:** `PROVISIONING_API_TOKEN` (new, min 32 chars — added to `fnox.toml`) and `AUTH_SECRET` (already required; salts the receipt and bucket ids).

## Auth model

- **Fails CLOSED on misconfiguration.** Unset, empty, whitespace-only or under-32-character secret refuses *everybody*. "No secret configured" never collapses into "no auth required".
- **Constant-time comparison.** Both sides are hashed to a fixed 32-byte digest before `crypto.timingSafeEqual`, so there is no length-dependent branch at all — stricter than the length-equalising `safeEqual` in `email-link/token.ts`, which has to preserve its callers' raw strings.
- **Token never echoed** into a response, an error or a log line. There is a test that greps the response body *and* the captured console output for both the presented and the configured secret.
- **Opacity is scoped to the unauthenticated.** Every auth failure returns a byte-identical 401, so a prober learns nothing about configuration, token shape, or whether a slug/domain exists (there is a test comparing a probe of a taken slug against a probe of a free one). An *authenticated* caller gets named conflicts — kontroll legitimately has to render "that slug is taken", and by then the caller is trusted platform infrastructure.
- **Two durable rate limits** (Sanity-backed CAS; an in-process counter is not a limit in serverless): a per-IP `attempt` bucket charged *before* the token is compared (10/min, 60/h, 300/day — bounds brute-force), and a *single global* `create` bucket charged after auth and validation (5/min, 30/h, 100/day — bounds bulk minting if the secret leaks, and is deliberately not keyed on anything a caller can rotate). Both **fail closed** in either direction, unlike the sign-in limiter which fails open on read outage.

The email-link bucket algorithm was extracted to `src/lib/rate-limit/` and both limiters now share it; the email-link suite passes unchanged, which is the proof the extraction is behaviour-preserving.

## Idempotency — what was chosen and why

**Both** an idempotency key *and* a uniqueness check inside the transaction, because they are the same mechanism here:

`Idempotency-Key` is required. It materialises as a `provisioningRequest` receipt at a deterministic id, `sha256(key + AUTH_SECRET)`, **created inside the same transaction as the organization**. Sanity's `create` on an explicit id fails if the document exists and the transaction is atomic — so a replayed key *cannot commit an organization*, however the two requests interleave. That is a genuine compare-and-swap; the pre-flight receipt read is only a fast path.

Why not rely on the existing slug-uniqueness read? Because it is a read-then-write check with a real TOCTOU window — two concurrent requests both see "slug free" and both commit. The sabotage below proves it: with the receipt removed, two racing identical requests create **two** organizations.

The receipt also makes the retry *safe* rather than merely refused: a caller whose response was lost gets the original ids back, so a timeout can never strand a tenant kontroll does not know about. The **key**, not the payload, is authoritative — a used key with a different body returns the original tenant and writes nothing.

Receipts are purged after 30 days by the daily cleanup cron, so retention *is* the replay window. Nothing about the key is stored — only its salted hash, which is already the document id, so a content-lake reader cannot confirm a guessed key.

**Residual risk, documented not hidden:** two requests with *different* keys racing on the *same* slug can still both pass the uniqueness read. The receipt is keyed on the request, not the slug. Closing it properly means deriving the organization `_id` from the slug so Sanity's create-CAS enforces it — a follow-up, called out in `docs/PROVISIONING_API.md`.

## Sabotage evidence

Baseline: **37 passed, 0 failed**. Each sabotage was applied to the real source, the suite re-run, and the source restored.

| # | Sabotage | Result | Representative failures |
| --- | --- | --- | --- |
| S1 | Unset secret means "no auth required" (`return { ok: true }`) | **4 failed / 33 passed** | all three unset/empty/whitespace cases assert `organization` count `0` and now see `1` |
| S2 | Secret compared on an 8-char prefix | **3 failed / 34 passed** | prefix token, extended token, and the bare-401 opacity case |
| S3 | Idempotency receipt not written in the transaction | **3 failed / 34 passed** | the race test fails on `expected [ {…}, {…} ] to have a length of 1 but got 2` — i.e. two organizations really were created |
| S4 | Payload not validated (raw body reaches the transaction) | **11 failed / 26 passed** | every malformed-input case plus the happy path |
| S5 | Pre-auth attempt limiter removed | **1 failed / 36 passed** | unauthenticated attempts are no longer metered |
| S6 | 401 body leaks the auth failure reason | **1 failed / 36 passed** | the bare-401 body assertion |
| S7 | 32-character secret floor removed | **1 failed / 36 passed** | a short secret is now honoured |
| S8 | `mintChallenges` allowed to throw post-commit (the pre-fix code) | **1 failed / 37 passed** | a committed tenant is reported as 500 and its ids are never returned |

Second round, after review flagged three documented properties with zero coverage. Baseline: **42 passed, 0 failed**. Each new test fails under its own sabotage and only its own, and each fails on a **document-count** assertion:

| # | Sabotage | Result | Assertion that went red |
| --- | --- | --- | --- |
| S9 | `onReadFailure: 'deny'` → `'allow'` | **1 failed / 41 passed** | `expected [ {…} ] to have a length of +0 but got 1` — a tenant was created during a store outage |
| S10 | A missing client IP skips the attempt limit | **1 failed / 41 passed** | `expected [] to have a length of 1 but got +0` — no attempt bucket was ever written |
| S11 | Creation bucket keyed on a caller-rotatable subject (`'global'` → random) | **1 failed / 41 passed** | `expected […8 items] to have a length of 5 but got 8` — IP rotation minted three extra tenants |
| S12 | `bucket.ts` returns `true` when a hit can never be persisted | **1 failed / 41 passed** | `expected [ {…} ] to have a length of +0 but got 1` — an unrecorded bucket let the write through |

**No production code changed for these** — all four behaviours were already correct (confirmed by live probe: no-IP requests gave `401×10` then `429×3`; IP-rotating creates gave `201×5` then `429×3` with exactly 5 orgs). They were unpinned invariants, not bugs. The Sanity fake now takes injectable read/write outages scoped to `provisioningRateLimit` documents, and the request helper can send no proxy headers at all.

Every assertion is on an **observable effect** — how many `organization` / `conference` / `speaker` documents ended up in the store and which ids came back. Error strings are only ever checked as a secondary property after the write count has been pinned. The in-memory Sanity fake models the three behaviours the design leans on: `create` on an existing `_id` fails, `patch().ifRevisionId()` fails on a moved revision, and a transaction rolls back in full.

## Gates

- `pnpm test` — 463 files, **6584 passed**
- `pnpm typecheck` — clean
- `npx eslint .` — **0 errors, 216 warnings** (baseline on this tree is also 0/216, verified by stashing)
- `npx prettier --check .` — clean
- `pnpm knip` — exit 0
- `pnpm build` — succeeds; `/api/provisioning/organizations` registers as a dynamic route

## Caught-and-fixed failure mode (`d1d2f33c`)

A real defect found during review, not a cosmetic follow-up.

`mintChallenges()` mints the DNS verification records and runs **after** `tx.commit()`. It was allowed to throw. So a Sanity read outage at that moment returned **`500` for a tenant that had actually been created** — and because the idempotency receipt's replay path calls the same helper, every subsequent retry threw too. The caller could therefore *never* learn the ids of its own committed tenant: the organization existed, held the domain claim, and was invisible to kontroll forever.

The fix catches both the sync and each per-domain read and logs them; a record that cannot be read simply projects as `pending`, which is where it already fails closed (never routed under enforcement, never allowlisted) and what the daily sweep re-mints. Pinned by a test and by sabotage S8 above — restoring the throwing version turns it red on `status 201` + the committed id.

## Notes for review

- Two new platform-internal Sanity types (`provisioningRequest`, `provisioningRateLimit`), both hidden from the Studio structure and both purged by the existing daily cron.
- `src/server/routers/onboarding.ts` behaviour is unchanged — its 22 existing tests pass untouched, which is the guard on the extraction.
- Deliberately **not merged or enqueued**; this wants an adversarial pass first.

<h3>Greptile Summary</h3>

The PR adds a bearer-authenticated, rate-limited, idempotent machine API for tenant creation and extracts the existing wizard transaction into a shared provisioning service.
- Adds provisioning authentication, durable attempt/create rate limits, idempotency receipts, schemas, cleanup, documentation, and integration tests.
- Reuses the shared organization, conference, and organizer transaction from both the tRPC wizard and the new HTTP route.
- A post-commit verification read can still make the new endpoint report failure after the tenant has been persisted.

The post-commit challenge failure must be handled before merging because the API can return an internal error after successfully creating and persisting a tenant.

The tenant transaction and idempotency receipt commit before an uncaught domain-verification fetch, allowing a transient read failure to produce a 500 that contradicts the persisted result.

**Files Needing Attention:** src/lib/onboarding/provision.ts and src/app/api/provisioning/organizations/route.ts

| Filename | Overview |
|----------|----------|
| src/app/api/provisioning/organizations/route.ts | Adds the authenticated machine API with uniform auth failures, validation, two rate-limit stages, conflict mapping, and shared provisioning invocation; its outer catch converts post-commit challenge failures into misleading 500 responses. |
| src/lib/onboarding/provision.ts | Centralizes tenant uniqueness checks and atomic writes with optional idempotency receipts, but performs a fallible verification read after commit without preserving the successful provisioning result. |
| src/lib/rate-limit/bucket.ts | Extracts the Sanity-backed rolling-window CAS algorithm and retains configurable read-failure behavior. |
| src/lib/provisioning/token.ts | Implements fail-closed bearer authentication with fixed-width digest comparison and no token reflection. |
| src/lib/provisioning/rateLimit.ts | Adds salted per-IP attempt and global post-authentication creation buckets with fail-closed persistence behavior. |
| src/server/routers/onboarding.ts | Replaces inline tenant creation with the shared provisioning service while retaining tRPC-specific rejection translation. |
| src/app/api/cron/cleanup-notifications/route.ts | Extends the existing daily cleanup route to purge expired provisioning receipts and rate-limit buckets. |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
  participant Kontroll
  participant Route as Provisioning route
  participant Limits as Sanity rate limits
  participant Provision as provisionOrganization
  participant Sanity
  participant Verify as Domain verification

  Kontroll->>Route: POST + bearer token + idempotency key
  Route->>Limits: Charge attempt bucket
  Route->>Route: Authenticate and validate
  Route->>Limits: Charge global create bucket
  Route->>Provision: Provision validated tenant
  Provision->>Sanity: Read receipt and uniqueness state
  Provision->>Sanity: Atomic tenant + receipt transaction
  Sanity-->>Provision: Commit succeeds
  Provision->>Verify: Sync and project challenges
  Verify--xProvision: Verification read can throw
  Provision--xRoute: Rejection after commit
  Route-->>Kontroll: 500 despite persisted tenant
```

<sub>Reviews (1): Last reviewed commit: ["feat(provisioning): bearer-authenticated..."](https://github.com/cloudnativebergen/website/commit/f4564134b05ae2e488ae3bbe097830d9500d9599) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=50153195)</sub>

**Context used:**


Starefossen:
**Adversarial review: SHIP.** Three invariants pinned since, and independently re-verified.

## What the review could not break

It attacked auth, replay, validation, domain overlap and the shared limiter, and every attack failed against correct code:

- **Fail-open on missing config** — refuses with the token unset, empty, whitespace, or under 32 chars: all `401`, zero documents written. The check runs *before* `request.json()`, so no body is parsed pre-auth.
- **Comparison** — both sides hashed to fixed 32-byte digests before `timingSafeEqual`, so the length-mismatch throw path does not exist at all.
- **Replay** — the receipt is created *inside* the transaction, making it a genuine CAS. Two concurrent identical requests produced exactly one org (`201` + `200`).
- **Information disclosure** — wrong-token probes against a taken vs. free slug returned byte-identical `401`s; auth precedes every content-lake read, so there is no timing channel.
- **Input validation** — injected `plan`, `entitlements` and `_id: "attacker-controlled-id"`; the written document had only `_id,_type,name,slug,contactEmail` with a generated id. Domain overlap refused in **both** directions.
- **Second implementation?** No. `onboarding.ts` deletes its copy and delegates, so the #752 enumeration defect is not forked.
- **The shared rate limiter** — this was my main concern, since it is used by email sign-in. Diffed line by line: a faithful extraction. Sign-in keeps its own failure directions, and buckets cannot collide (different `_type`, id prefix, and hash domain separation). **Sign-in rate limiting is not weakened.**

## What it found, and what changed

Three documented security properties had **zero coverage** — sabotage each and all 38 tests still passed: `onReadFailure: 'deny'`, a missing client IP charging the shared bucket rather than skipping the limit, and the global create bucket being unrotatable.

The code was correct; the invariants were unpinned. That is the failure mode where a future refactor deletes a guarantee silently, so they are now tested (42 tests, up from 38).

**Verified myself rather than taking it on trust** — I made the global bucket caller-rotatable, the single control the constants file says must not be evadable by rotating `x-forwarded-for`:

| state | result |
| --- | --- |
| unrotatable | 42/42 pass |
| made rotatable | **1 failed** — `caps creation GLOBALLY — rotating the client IP does not buy more tenants` |
| restored | 42/42 pass |

Also worth naming: `d1d2f33c` fixed a real defect caught mid-review. A throw in `mintChallenges()` — which runs *after* the transaction commits — returned `500` for a tenant that **had been created**, and since the replay path calls it too, every retry threw again, so a caller could never learn the ids of its own tenant.

One follow-up filed separately: the pre-auth attempt limiter creates a document per client IP, so a rotating `x-forwarded-for` writes documents without a valid token. Pre-existing platform pattern (the email-link limiter does the same), not a regression, but worth a cheap global pre-auth bucket.

Issue #688, opened 2026-07-28
Central auth origin: allowlist state.origin against verified domains before bouncing the OAuth code


Starefossen:
Required before serving auth on tenant-owned domains via the central auth origin (`auth.konf.app`).

## The gap
With `AUTH_REDIRECT_PROXY_URL` enabled, the proxy callback in `@auth/core` (`lib/actions/callback/index.js:25-35`) validates only that:
1. the `state` parameter decrypts (JWE, using `AUTH_SECRET`), and
2. the decoded origin differs from the proxy's own origin.

**There is no allowlist of permitted destinations.** The only thing standing between an arbitrary origin and a real authorization code is possession of `AUTH_SECRET`.

Concrete failure: if preview deployments share the production `AUTH_SECRET`, a preview host can mint a `state` naming itself, and the production proxy will forward a genuine authorization code to it. The same applies to any future self-hosted instance sharing the secret.

**Note the correction:** this is *not* mitigated by PKCE. LinkedIn's Auth.js provider ships `checks: ["state"]` only, and GitHub's OAuth server does not enforce PKCE — so `state` is the *only* binding across both configured providers. That makes the allowlist **mandatory, not defence-in-depth**.

## Requirements
- Wrap the proxy callback so `state.origin` is checked against an allowlist **before** bouncing.
- **EXACT-STRING match** against **ownership-verified** domains (see #683). Per RFC 6819 §4.1.5/§5.2.3.5 and RFC 9700 §2.1/§4.1.3/§4.11.1, prefix or wildcard matching combined with any open redirector is the canonical code-exfiltration chain.
- ⚠️ **Do not reuse the routing matcher.** Conference routing *deliberately* matches `*.example.com` wildcards (`domainEntriesOverlap`, #666/#679/#681). The redirect allowlist must **not** inherit that behaviour — same data, deliberately different matching rules.
- Unverified domains are excluded from the allowlist even if present in `domains[]`.
- Fail closed: unknown origin → reject, do not bounce.
- Pair with distinct `AUTH_SECRET` per environment (operational, tracked separately).

## Acceptance
- A `state` naming an origin not in the verified allowlist is refused, and no code is forwarded.
- Tests cover: exact verified host accepted; unverified-but-claimed host refused; wildcard-adjacent host refused (proving the routing matcher was not reused); preview origin refused.
- Verify the tests bite by removing the allowlist check.

Starefossen:
[...]

PR #778, opened 2026-08-04
feat(domains): platform-owned verification for hosts in our own zone

Starefossen:
## The problem

Tenants are hosted by default on a subdomain the platform **mints** — `<slug>.konf.run`. That zone's nameservers are delegated to our own edge and a `*.konf.run` wildcard domain is registered on this Vercel project, so we control it end to end and no tenant has any access to it.

`src/lib/domain-verification/` had no notion of that. `isRoutingEligible` decided purely from a `domainVerification` record, and the only non-DNS method was `grandfathered` — a deliberately **time-boxed** 30-day exemption. So the moment `DOMAIN_VERIFICATION_ENFORCE_ROUTING=true` is set, every platform-hosted tenant either stops routing or is asked to publish a `_konf-challenge` TXT record in a zone only we can write to.

## The method: an ALLOCATION, not an inference

> **Review correction.** The first cut of this PR inferred entitlement from the SUFFIX at read time — any host under `PLATFORM_DOMAIN_SUFFIX` was verified. That reasoning is correct about *why* DNS-TXT proof is impossible in our own zone, but it does not follow that the claiming tenant is entitled to the hostname. An organizer can type any hostname into /admin/settings, so they could enter `some-other-tenant.konf.run` — a label the platform never issued, or one earmarked for a customer being onboarded next week — and be handed routing plus (once #688 ships) an OAuth redirect destination. Global uniqueness makes that *worse*: the grab is exclusive, so the rightful tenant could then never be given the hostname at all. The model below is the fix.

**Being under the suffix is a PRECONDITION. The entitlement is a grant recorded at write time.**

- The platform grants one host to one conference through `provisionOrganization` — the platform-operator wizard and the bearer-authenticated provisioning API. It is the **only** caller that passes `allocatePlatformHosts`, and it is unreachable by a tenant organizer.
- The `domainVerification` document records the grant: `method: "platform-owned"`, `conference` = the grantee.
- Every read-time decision calls `isPlatformAllocated`, which requires **both** halves — the recorded grant **and** a live suffix re-check. A mis-issued record for a host outside the zone grants nothing; a host in the zone with no grant grants nothing; re-pointing or unsetting the suffix withdraws the standing immediately.

Once allocated the standing is **permanent** — no `graceUntil`, no staleness expiry, nothing for the tenant to complete. Revocation, not expiry, is how it ends.

### Defence in depth at the write path

| Layer | Behaviour |
| --- | --- |
| `updateDomains` / `createEdition` | **Reject** the payload (`findUnallocatedPlatformDomains` → BAD_REQUEST) before any write. The claim itself is never made — which matters, because `domains[]` is globally unique and a saved grab would lock the rightful tenant out permanently. |
| `ensureDomainVerification` | Writes **no document at all** for an in-zone host without the allocation flag, so a bypass of the mutation guard still fails closed. It also refuses to re-allocate a host held by another conference, and an explicit `method: 'platform-owned'` argument cannot smuggle a grant in. |
| `isHostRoutable` | The suffix short-circuit is **gone**. A missing record fails closed for platform and custom hosts alike. |
| The sweep | Keys its no-lookup path on the allocation. An unallocated in-zone claim is resolved like any other, hard-fails, and stays unrouted. |
| `toDomainVerificationView` | Record-driven, not hostname-driven, and no longer forces `status: verified` / `routable: true`. The card never announces "provided by the platform" for a host the platform did not issue. |

A tenant that releases and re-adds *its own* allocated subdomain is restored without a support ticket — keyed on the stored grant, so it can only ever restore one the platform actually made.

## How the suffix is configured

`PLATFORM_DOMAIN_SUFFIX`, following the existing `PLATFORM_ORG_SLUG` contract (`src/lib/features/platform.ts`) — the platform is white-labelable, so `konf.run` is a deployment fact, never a constant in the source. Documented in `.env` next to `PLATFORM_ORG_SLUG`.

**Unset means _no_ host is in the platform zone**, never "everything is". Every rejection path returns `null` and `isPlatformZoneHost` fails closed on it:

| Configured value | Result |
| --- | --- |
| unset / `""` / `"   "` | `null` — nothing is allocatable |
| `run` (single label) | `null` — would otherwise cover every `.run` domain |
| `https://konf.run`, `konf.run/path`, `konf.run:3000`, `konf run` | `null` |
| `konf.run`, `.konf.run`, `*.konf.run`, `  KONF.run ` | `konf.run` |

## How the suffix is matched

Both sides are split into DNS **labels**; the suffix must be the exact trailing labels of the host with at least one label to spare. No `endsWith`, no substring logic, and normalisation goes through the repo's own `normalizeDomain` / `isValidDomainEntry`. "✅" below means only *allocatable* — the grant still has to exist.

| Host | vs `konf.run` | Why |
| --- | --- | --- |
| `kubeday.konf.run` | ✅ in zone | exact trailing labels, one deeper |
| `a.b.konf.run` | ✅ in zone | still inside the zone |
| `evil-konf.run` | ❌ | `'evil-konf.run'.endsWith('konf.run')` is **true** — there is no label boundary |
| `konf.run.attacker.com` | ❌ | our zone as somebody else's *prefix* |
| `a.konf.runner`, `a.konf.ru` | ❌ | different TLD |
| `konf.run` (apex) | ❌ | the platform's own origin, not a minted subdomain |
| `*.konf.run` | ❌ | a wildcard over the whole zone would hand its holder every tenant subdomain |
| `tenant.konf.run:3000`, `tenant.konf.run.` | ❌ | a port is a dev entry; a trailing dot is not a label |

## Redirect allowlist: included — and why

`isAllowlistEligible` honours an **allocated** host. The threat it guards is a *dangling* destination: a third party's zone lapses and the host silently resolves to somebody else, which the victim experiences as a normal login. That cannot happen inside a zone we operate — its delegation cannot change without our own registrar account changing hands, and not quietly, because every tenant site would go down at once. The 30-day staleness rule exists to catch a silently broken checker; for our own zone there is no checker to break. Excluding these would mean nobody on the platform's default hosting could complete a sign-in.

What it grants is a redirect destination to the conference the platform **issued** that subdomain to — not to whoever typed it first. Wildcards stay excluded (RFC 9700 §4.1.3 untouched), dev hosts stay excluded, and `revoked` is checked **before** the allocation everywhere, so releasing the claim removes the grant instantly.

## Other review fixes

- **Revoked wins** over the allocation in routing, the allowlist and the view. Revocation is the remedy if a host ever ends up with the wrong tenant, so it has to actually undo the grant.
- **`graceUntil` can be cleared.** `DomainVerificationPatch` now carries the field, so a record that was grandfathered before it was allocated has its 30-day deadline unset instead of expiring a grant that has no expiry.

## Sabotage evidence

Baseline: **239/239** across `src/lib/domain-verification` + `conference.test.ts` (full suite 6694/6694).

| # | Sabotage | Observed |
| --- | --- | --- |
| 1 | Entitlement inferred from the suffix again — policy grants on zone membership, `ensureDomainVerification` allocates implicitly, the mutation guard withholds nothing, the view announces it | **15 failed** / 224 passed |
| 2 | `revoked` loses to the allocation (checked after it in the policy; dropped from the view) | **4 failed** / 235 passed |
| 3 | Stale `graceUntil` left in place (dropped from the check write-back and from the in-place upgrade's `unset`) | **3 failed** / 236 passed |

Earlier, still-passing controls from the first round, re-verified: making the matcher a naive `endsWith` fails 5 (`evil-konf.run` and the apex), and making an unset suffix mean "match everything" fails 23.

Every assertion is on **observable behaviour** — whether routing is granted, whether the host is on the allowlist, whether a DNS lookup was issued, whether a document was written and with what — never on an error message.

Named failures under sabotage 1 include `REFUSES an UNALLOCATED host that merely sits in the platform zone` (policy **and** allowlist), `FAILS CLOSED for an in-zone host with NO record — no suffix exemption` (routing), `writes NOTHING for an in-zone host when the caller may not allocate` and `will not let an explicit method smuggle an allocation in` (persistence), `REFUSES an in-zone host with no record — the unissued-label grab` and `REFUSES a host allocated to ANOTHER conference` (the mutation guard), `DNS-CHECKS an in-zone host the platform never allocated` (sweep), and `does NOT announce an in-zone host the platform never allocated` (view).

## Custom domains are unaffected

Proved directly rather than by omission — `leaves CUSTOM domains exactly as they were — real proof still required` (policy), `leaves CUSTOM domains fail-closed on a missing record` (routing), `REFUSES an unproven CUSTOM domain while the platform rule is on` (allowlist), `still delists a CUSTOM domain swept alongside an allocated one` (sweep), `IGNORES custom domains entirely` (the mutation guard), `mints a CUSTOM domain pending/dns-txt` (persistence).

## Gates

`pnpm test` 6694/6694 · `pnpm typecheck` clean · `npx eslint .` **0 errors, 216 warnings (unchanged)** · `npx prettier --check .` clean · `pnpm knip` exit 0 · `pnpm build` exit 0.

Visual inspection per AGENTS.md: `PlatformOwned` story plus a row in `AllStates`, captured with `pnpm shoot` at 393px — no overflow, consistent with the existing states.

## Rollout note

`PLATFORM_DOMAIN_SUFFIX=konf.run` must be set in the Vercel environment **before** `DOMAIN_VERIFICATION_ENFORCE_ROUTING=true` is switched on, and existing `konf.run` tenants must be allocated (the backfill deliberately does **not** allocate — it cannot know whether the platform ever issued a label to a conference). Until the suffix is set this change is a complete no-op: no host is in the platform zone and every code path behaves exactly as it does today.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01N1XZpKx5SjpEPdT9sLYrY2

### DIFF
--- a/.env
+++ b/.env
@@ -16,11 +16,15 @@ NEXT_TELEMETRY_DISABLED=1
 # PLATFORM_ORG_SLUG="cloud-native-bergen"
 
 # The zone the PLATFORM itself operates and mints tenant subdomains in
-# (`<slug>.konf.run`). Any host strictly BELOW this suffix is treated as
-# ownership-verified by construction (`method: "platform-owned"`) — permanently,
-# not on a grace period — because its DNS is ours and no tenant could ever
-# publish the challenge record there. See docs/DOMAIN_VERIFICATION.md.
-# Unset = NO host is platform-owned (fail closed). It is never "match
+# (`<slug>.konf.run`). A host strictly BELOW this suffix is ALLOCATABLE: tenant
+# provisioning may grant it to one conference, and the resulting record
+# (`method: "platform-owned"`) is then ownership-verified by construction —
+# permanently, not on a grace period — because its DNS is ours and no tenant
+# could ever publish the challenge record there.
+# Being under the suffix is NOT by itself an entitlement: an unallocated host is
+# refused when an organizer tries to claim it, and grants nothing if one ever
+# slips through. See docs/DOMAIN_VERIFICATION.md.
+# Unset = NO host is in the platform zone (fail closed). It is never "match
 # everything". Matching is label-wise: `evil-konf.run` and
 # `konf.run.attacker.com` do NOT match `konf.run`, and neither does the apex nor
 # a `*.konf.run` wildcard claim. Custom domains are unaffected either way.
--- a/.env
+++ b/.env
@@ -15,6 +15,18 @@ NEXT_TELEMETRY_DISABLED=1
 # Set it in .env.local / Vercel, e.g.:
 # PLATFORM_ORG_SLUG="cloud-native-bergen"
 
+# The zone the PLATFORM itself operates and mints tenant subdomains in
+# (`<slug>.konf.run`). Any host strictly BELOW this suffix is treated as
+# ownership-verified by construction (`method: "platform-owned"`) — permanently,
+# not on a grace period — because its DNS is ours and no tenant could ever
+# publish the challenge record there. See docs/DOMAIN_VERIFICATION.md.
+# Unset = NO host is platform-owned (fail closed). It is never "match
+# everything". Matching is label-wise: `evil-konf.run` and
+# `konf.run.attacker.com` do NOT match `konf.run`, and neither does the apex nor
+# a `*.konf.run` wildcard claim. Custom domains are unaffected either way.
+# Set it in .env.local / Vercel, e.g.:
+# PLATFORM_DOMAIN_SUFFIX="konf.run"
+
 NEXT_PUBLIC_SANITY_PROJECT_ID="mvzwvw14"
 NEXT_PUBLIC_SANITY_DATASET="production"
 SANITY_STUDIO_PROJECT_ID="mvzwvw14"
--- a/docs/DOMAIN_VERIFICATION.md
+++ b/docs/DOMAIN_VERIFICATION.md
@@ -47,7 +47,7 @@ A `*.example.com` wildcard claim is proven on its **base zone** (`example.com`).
   defending against, is the attacker. A record in the tenant's own zone proves
   control of the **zone**, which is the thing that actually lapses.
 
-## Platform-owned hosts
+## Platform-allocated hosts
 
 Tenants are hosted by default on a subdomain the platform **mints** for them
 (`<slug>.konf.run`). That zone's nameservers are delegated to our own edge and a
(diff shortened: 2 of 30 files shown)


Starefossen:
**The hijack finding was right, and the mechanism has been redesigned.** Verified independently before merge.

## What was wrong

`platform-owned` inferred ownership from the **suffix at read time**: any host under `PLATFORM_DOMAIN_SUFFIX` was treated as verified. The reasoning — that no tenant can publish a challenge record in a zone we control — is correct about why DNS proof is impossible, but it does not support the conclusion. **"This host is in our zone" does not imply "this tenant is entitled to it."**

An organizer adding a domain through their own settings could enter `some-other-tenant.konf.run`, a label the platform had not allocated, and receive routing plus — once #688 ships — an OAuth redirect destination. Global uniqueness made it worse, not better: the claim becomes exclusive, so the legitimate tenant could never be given that hostname.

## What changed

Entitlement is now an **allocation recorded at write time**: `method === 'platform-owned' && isPlatformZoneHost(...)`. Being in our zone is a precondition, never sufficient. The write paths reject an unallocated in-zone host rather than silently verifying it at read.

## Verified myself

Reducing the check back to suffix-only — the original vulnerability exactly — fails **6 tests**, each named for the property it defends:

> `REFUSES an UNALLOCATED host that merely sits in the platform zone` · `does NOT announce an in-zone host the platform never allocated` · `FAILS CLOSED for an in-zone host whose record is merely pending` · `REFUSES an in-zone host that a tenant merely got VERIFIED by DNS` · `DNS-CHECKS an in-zone host the platform never allocated`

Restored: 144/144 pass (up from 114).

Also confirmed from the earlier round, which I had verified before the redesign and which still holds: replacing the label-boundary matcher with a naive `endsWith` fails on `evil-konf.run` and on the apex; making an unset suffix mean "match everything" fails six tests across verification, allowlist and enforcement. Fail-closed in both directions.

The other three findings are addressed too: `DomainVerificationPatch` now carries `graceUntil: string | null` so a stale grandfathered deadline can be cleared, the platform-owned branch sets it to `null`, and revoked is refused before the allocation check.

## One narrow gap, pre-existing — filed separately

Removing the `revoked` guard inside `inGrandfatherGrace` passes 144/144. The case needs a record that is grandfathered **and** revoked **and** still inside its 30-day window, so today that is one of three Cloud Native domains after an admin revocation. `main` already has this guard, so it is not a regression from this PR — but it is another invariant no test observes.

PR #779, opened 2026-08-04
fix(email): send unverified tenants from the platform sender, never silently

Starefossen:
Closes the send half of RunKonf/platform#20. A newly provisioned tenant could not receive email, and the failure was invisible.

## ⚠️ Operational gate: the platform cannot currently send as itself

**`konf.app` is NOT verified in Resend** — no `resend._domainkey.konf.app`, no `send.konf.app` MX. The only domains carrying Resend DKIM + `send.` MX today are **`cloudnativedays.no`** and **`cloudnativebergen.dev`**, and *both belong to tenants*. `EMAIL_FALLBACK_FROM` is also **unset** in the `konf` production project.

So until `konf.app` is verified, the platform sender would have to **borrow a tenant's domain** (`EMAIL_FALLBACK_FROM="Konf <noreply@cloudnativedays.no>"`), which works but is not what we want to ship long-term. `konf.app` is on **registrar DNS (Namecheap), not Vercel**, so verifying it is a **manual step for the domain owner** — it cannot be automated from this repo.

**Follow-up, required before tenant #2 sends email:**
1. Verify `konf.app` in Resend (add its DKIM/SPF records at Namecheap).
2. Set `EMAIL_FALLBACK_FROM="Konf <noreply@konf.app>"`.
3. Set `EMAIL_SENDING_DOMAINS` to every verified domain, so tenant #1 keeps sending as itself.

Until step 2 the new **Effective senders** check reads `error` and says exactly this.

## The bug, confirmed

`resolveConferenceFrom` picks a tenant-shaped `From:` — `contactEmail`, else `<localPart>@<the conference's own domain>`. On the shared tier that message is sent on the **platform's** Resend account, which may only send from domains verified for **that** account. Resend rejects it. `requestEmailSignInLink` then returns its deliberately uniform anti-enumeration outcome, so the tenant cannot sign in and nothing surfaces. Masked so far because tenant #1's domain is verified.

Three things the original analysis did not cover, found while establishing the truth:

- **The helper is not the only offender.** Roughly half the ~20 `resend.emails.send(...)` call sites never go through `resolveConferenceFrom` — they build `` `${conference.organizer} <${conference.cfpEmail}>` `` inline (`email/speaker.ts`, `email/gallery.ts`, `messaging/email.ts`, `proposal/email/notification.ts`, …). A policy applied only inside the helper would have missed them.
- **`email/badge.ts` had its own `new Resend(process.env.RESEND_API_KEY)`**, bypassing the shared client entirely. It is now routed through it.
- **`resend.broadcasts.create` is a second send API** with its own `from`, used by the audience broadcast path. Also now instrumented.

## The sender policy

Identity and envelope are now separate questions:

```
From:     KCD Bergen <noreply@[platform sending domain]>
Reply-To: hello@kcd.dev
```

The recipient still sees the conference; replies still reach the organizers; the message is one the platform account is allowed to send.

| Case | Behaviour |
| --- | --- |
| Tenant domain **not** verified | Rewritten to `EMAIL_FALLBACK_FROM`, tenant name kept, tenant address → `Reply-To` |
| Tenant domain in `EMAIL_SENDING_DOMAINS` | Untouched — sends as itself |
| Tenant's **own** Resend account | Exempt — its domains are verified on its own account |
| `EMAIL_FALLBACK_FROM` unset | Passed through (nothing deliverable to rewrite to), logged once per domain, **error** on `/admin/settings` |

That last row is deliberate: rewriting to a placeholder would break the tenant that works today instead of just the one that doesn't.

## A conference has THREE senders, not one

`contactEmail`, `cfpEmail` and `sponsorEmail` can sit on **different domains**, and each carries its own flows:

| Field | Sends |
| --- | --- |
| `contactEmail` | sign-in links, badges, invitation letters, workshops, volunteers, broadcasts, proposal notifications |
| `cfpEmail` | speaker mail, speaker/sponsor messaging, gallery, co-speaker invites |
| `sponsorEmail` | sponsor CRM + registration, contract signing and reminders |

The first revision of the **Effective senders** check judged only `contactEmail` — it could have reported health while CFP mail was being rejected. A diagnostic that lies is worse than no diagnostic, because it is read precisely when something is already wrong. It now evaluates all three, takes the worst verdict, and **names the offending address**; where some senders still work it says so explicitly, so a partially-green reading cannot be mistaken for health.

The **Send test email** probe had the mirror-image blind spot — it always sent from `cfpEmail`, so it could come back green off a healthy address while another was refused. It now sends from the conference's **worst** sender and reports which one it used.

The inventory (`CONFERENCE_SENDER_FIELDS`) is derived from the **send sites**, not the schema, precisely because several flows build the header inline.

## Header injection, fixed at the source

`parseAddress()` sanitised the display **name** but returned the address **raw** in both branches, and `applySenderPolicy` derives `Reply-To:` from `.address`. `From:` headers are built from tenant-editable conference fields and ~half the send sites interpolate them raw, so an organizer storing `hello@kcd.dev\r\nBcc: attacker@evil.example` got those bytes into a header. Authenticated-organizer trust level, multi-tenant platform — the blast radius is other tenants' recipients.

Fixed at the **client boundary**, not inside the policy. The first attempt put it in `parseAddress` — but that is a function the *policy* path happens to call, and **dedicated clients (the `dedicated-email` Pro entitlement) skip the policy by design**, so the injection stayed open for exactly the tenants who pay for their own Resend account. Same fields, same raw-interpolating send sites, a branch the policy never touches.

So `instrumentResendClient` now sanitises `from` and `replyTo` on **every send, on every client**, whichever policy branch ran. The `parseAddress`/`applySenderPolicy` sanitisation stays as defence in depth.

**Sanitisation truncates at the first CR/LF rather than deleting the breaks.** Deletion splices the payload onto the value, and for an address that hands the attacker the resulting *domain*: `a@evil.example\r\nb@verified.test` would collapse to a single address whose domain reads as verified, defeating the sending-domain check outright. Truncation keeps the legitimate prefix and discards the payload. One shared sanitiser, imported by `email/from` instead of redefined there.

**Every `from`-bearing Resend method is wrapped** — `emails.send`/`create`, `batch.send`/`create`, `broadcasts.create`/`update` — including the ones this codebase does not call today, so adopting one later cannot silently reopen the hole. (`broadcasts.send` takes only an id.)

**Is the instrumented client the only path out?** Yes, established three ways: `new Resend(` appears only in the client factory (`email/config.ts`) — `email/badge.ts` had its own and was routed through the shared client earlier in this PR; there are no direct `api.resend.com` calls; and the SDK surface was enumerated for `from`-carrying payloads, which is how `batch` and `broadcasts.update` were found.

Swept the other consumers — `addressDomain`, `platformSendingDomains`, `describeSenderPolicy.requested`, `conferenceSenders().address`, the probe's `sentAs` — all clean. `resolveConferenceContact` reaches JSON (badge issuer), an `href`, and email *body* copy rather than any SMTP header, so it was never an injection path; sanitised anyway so a future header consumer is not the next hole.

Assertions are **structural** — no CR, no LF, `split(/\r\n|\r|\n/)` length 1, and no `bcc` property on the constructed payload. Not substring checks: sanitised output can legitimately contain the payload text, so `not.toContain('Bcc:')` would be both false-negative and false-positive prone.

## The Pro-tier seam (not built here)

`platformSendingDomains()` is the single place that answers "may the platform account send as this domain?". Making it self-service is replacing its body with a cached `resend.domains.list()` lookup filtered to `status === 'verified'` — no call site moves. It additionally needs per-tenant DNS instructions in the admin UI, a verification poller, and cache invalidation, because a live API call must never sit in a send path. The other path — a dedicated Resend account per org — already exists via `TENANT_SECRETS_JSON`; gating it on the `dedicated-email` Pro entitlement rather than on the presence of a secret remains RunKonf/platform#26.

## Never silent again

The policy and the failure logging live on the **client** (`instrumentResendClient`), not on the call sites — so a send path that forgets, or swallows its error into a boolean nobody reads, is still covered:

```
[email] send failed {
  orgId, from, replyTo, recipientDomains: ['example.com'],
  senderPolicy: 'platform-rewritten',
  error: { name: 'validation_error', message: '…' }
}
```

The anti-enumeration property is untouched: the user-facing outcome is identical in every branch. Observability is entirely operator-side — this log, the `[email-link]` log that adds the conference name, the **Effective senders** check, and the **Send test email** self-check. Recipients are logged as **domains only** and subjects never, since a sign-in mail's recipient is precisely what the design refuses to disclose.

## Sabotage evidence

Baseline: **472 files / 6745 tests pass**.

**(a) Restore the tenant-domain `From:`** — reverted the rewrite in `applySenderPolicy` to `from: message.from`:

```
Test Files  3 failed | 469 passed (472)
Tests       8 failed | 6715 passed (6723)
```

Failing: `sender-policy.test.ts` ×4 (rewrite, borrowed display name, explicit reply-to preserved, provisioned-tenant end-to-end), `instrument.test.ts` ×2 (`send` and the `create` alias), `email-link-send.test.ts` ×2 (the sign-in message, and the log context).

**(b) Swallow a send failure without logging** — deleted both `logEmailSendFailure` calls from the client wrapper:

```
Test Files  2 failed | 470 passed (472)
Tests       5 failed | 6718 passed (6723)
```

Failing: `instrument.test.ts` ×4 (returned `{error}`, thrown error, the no-PII assertion, the dedicated-account case) and `email-link-send.test.ts` ×1.

**(c) Judge only `contactEmail` again** — `conferenceSenders(conference).slice(0, 1)` in the check, against a conference whose `contactEmail` is on a verified domain and whose `cfpEmail`/`sponsorEmail` are not:

```
Test Files  1 failed | 471 passed (472)
Tests       3 failed | 6726 passed (6729)
```

Failing: the REJECTED case, the rewritten case, and the per-field domain-derivation case — each asserts the offending address is named.

**(d) Probe back to a fixed sender** — `const worst = senders[1]` (always CFP):

```
Tests  1 failed | 9 passed (10)   [status.probes.test.ts]
```

**(e) Revert the address sanitisation in `parseAddress`** (both branches back to `.trim()`):

```
Test Files  2 failed | 470 passed (472)
Tests       8 failed | 6733 passed (6741)
```

Failing: both `parseAddress` branch tests, the domain-confusion test, the derived-`Reply-To` and pass-through-`From` cases, the `resolveConferenceFrom`/`resolveConferenceContact` path, and two on the constructed message in `instrument.test.ts` (send + broadcast).

**(f) Echo the caller's header in the pass-through branches** (`from: message.from`):

```
Test Files  2 failed | 470 passed (472)
Tests       3 failed | 6738 passed (6741)
```

Failing: the verified-domain and unconfigured pass-through cases, plus the constructed-message assertion on the verified path — confirming that layer is pinned independently of (e).

**(g) Remove the client-level sanitisation, leaving `applySenderPolicy` intact** — the control for "is the new guard load-bearing or shadowed by the old one?":

```
Test Files  1 failed | 471 passed (472)
Tests       4 failed | 6741 passed (6745)
```

Failing: **only** the four DEDICATED-path tests (client send, caller-supplied `Reply-To`, broadcast, batch + broadcast `update`). Every shared-path injection test still passed — which is precisely the point: the policy covers the shared path, and nothing but this guard covers the dedicated one.

**(h) Swap truncation for deletion** in `sanitizeHeaderText`:

```
Test Files  2 failed | 470 passed (472)
Tests       5 failed | 6740 passed (6745)
```

Failing: `truncation denies the attacker the resulting DOMAIN`, the display-name truncation test, the verified-branch pass-through, and two dedicated-path constructed-message tests.

Assertions are on the **constructed message** (`sent[0].from`, `sent[0].replyTo`, the probe's `from`) and on **observed `console.error` calls with their context objects** — never on error strings.

## Gates

`pnpm test` 6745 pass · `pnpm typecheck` clean · `npx eslint .` 0 errors / 216 warnings (identical to `origin/main`, verified by stashing) · `npx prettier --check .` clean · `pnpm knip` only the pre-existing `@workos-inc/node` config hint · `pnpm build` succeeds.

Touches no file in `src/lib/onboarding/` or `src/lib/domain-verification/`.

### DIFF
--- a/.env
+++ b/.env
@@ -31,6 +31,24 @@ NEXT_TELEMETRY_DISABLED=1
 # Set it in .env.local / Vercel, e.g.:
 # PLATFORM_DOMAIN_SUFFIX="konf.run"
 
+# THE PLATFORM SENDER (RunKonf/platform#20). Outbound mail is addressed FROM the
+# conference's own identity, but it is SENT on the platform's Resend account —
+# which may only send from domains verified for that account. A tenant whose
+# domain is not verified therefore sends from this address instead, carrying the
+# conference name as the display name and its own address as Reply-To.
+# The address MUST be on a domain verified in the platform Resend account.
+# Unset = there is nothing deliverable to fall back to: such a tenant's mail is
+# REJECTED by Resend. That is logged and shown as an error on /admin/settings,
+# never papered over. Set it in .env.local / Vercel, e.g.:
+# EMAIL_FALLBACK_FROM="Konf <noreply@example.com>"
+
+# Every domain verified on the platform Resend account, comma-separated. A
+# conference whose From lands on one of these sends as ITSELF — this is how a
+# tenant graduates to its own sender once its DKIM/SPF records verify. The
+# EMAIL_FALLBACK_FROM domain is always included implicitly. Unset = only that
+# one domain. See docs/EMAIL_SYSTEM.md → "The sender policy".
+# EMAIL_SENDING_DOMAINS="example.com,another-verified.dev"
+
 NEXT_PUBLIC_SANITY_PROJECT_ID="mvzwvw14"
 NEXT_PUBLIC_SANITY_DATASET="production"
 SANITY_STUDIO_PROJECT_ID="mvzwvw14"
--- a/__tests__/lib/auth/email-link-send.test.ts
+++ b/__tests__/lib/auth/email-link-send.test.ts
@@ -1,0 +1,157 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest'
+import type { Resend } from 'resend'
+
+/**
+ * THE BUG, AT THE PLACE IT BIT (RunKonf/platform#20).
+ *
+ * A newly provisioned tenant's sign-in mail was addressed `From:` its own,
+ * unverified domain and sent on the PLATFORM's Resend key. Resend refused it,
+ * and `requestEmailSignInLink` returns the same opaque outcome either way — so
+ * the tenant simply could not log in, and nothing said so.
+ *
+ * These tests pin both halves of the fix on the real send path: the message that
+ * leaves carries a platform-verified `From:` with the tenant in `Reply-To:`, and
+ * a rejected send is on the record with the conference named.
+ */
+
+const { mockResolveEmailSender } = vi.hoisted(() => ({
+  mockResolveEmailSender: vi.fn(),
+}))
+
+vi.mock('@/lib/email/config', async (importOriginal) => {
+  const actual = await importOriginal<typeof import('@/lib/email/config')>()
+  return { ...actual, resolveEmailSender: mockResolveEmailSender }
+})
+
+import { instrumentResendClient } from '@/lib/email/instrument'
+import { resetSenderPolicyWarnings } from '@/lib/email/sender-policy'
+import { sendEmailSignInLink } from '@/lib/auth/email-link/send'
+import type { Conference } from '@/lib/conference/types'
+
+const PLATFORM_FROM = 'Konf <noreply@platform.example>'
+
+/** A tenant provisioned on its own domain — unverified on the platform account. */
+const TENANT = {
+  title: 'KCD Bergen 2026',
+  organizer: 'KCD Bergen',
+  contactEmail: 'hello@kcd.dev',
+  domains: ['2026.kcd.dev'],
+  city: 'Bergen',
+  country: 'Norway',
+  startDate: '2026-09-01',
+  socialLinks: [],
+} as unknown as Conference
+
+interface SentMessage {
+  from: string
+  replyTo?: string | string[]
+  to: string
+  subject: string
+}
+
+function stubSender(result: { data?: unknown; error?: unknown }) {
+  const sent: SentMessage[] = []
+  const send = vi.fn(async (payload: SentMessage) => {
+    sent.push(payload)
+    return result
+  })
+  // Shaped like the real thing: the PLATFORM client is shared by every tenant,
+  // so it carries no org of its own — the tenant is identified by the sender.
+  const client = instrumentResendClient(
+    { emails: { send, create: send } } as unknown as Resend,
+    { enforceSenderPolicy: true },
+  )
+  mockResolveEmailSender.mockResolvedValue({ client })
+  return sent
+}
+
+const REQUEST = {
+  to: 'speaker@example.com',
+  signInUrl: 'https://2026.kcd.dev/auth/email-link?token=secret-token-value',
+  expiresInMinutes: 15,
+  singleUse: true,
+  conference: TENANT,
+  orgId: 'org-kcd',
+}
+
+beforeEach(() => {
+  resetSenderPolicyWarnings()
+  mockResolveEmailSender.mockReset()
+  vi.stubEnv('EMAIL_FALLBACK_FROM', PLATFORM_FROM)
+  vi.stubEnv('EMAIL_SENDING_DOMAINS', '')
+})
+
+afterEach(() => {
+  vi.unstubAllEnvs()
+  vi.restoreAllMocks()
+})
+
+describe('sendEmailSignInLink for a tenant whose own domain is unverified', () => {
+  it('sends from the platform-verified sender, with the tenant address as Reply-To', async () => {
+    const sent = stubSender({ data: { id: 'email_1' } })
+
+    await expect(sendEmailSignInLink(REQUEST)).resolves.toBe(true)
+
+    expect(sent).toHaveLength(1)
+    expect(sent[0].from).toBe('KCD Bergen <noreply@platform.example>')
+    expect(sent[0].replyTo).toBe('hello@kcd.dev')
+    // The tenant's unverified domain must not appear in the envelope sender —
+    // that is exactly the header Resend rejects.
+    expect(sent[0].from).not.toContain('kcd.dev')
+    // The recipient still sees who it is from.
+    expect(sent[0].from).toContain('KCD Bergen')
+    expect(sent[0].subject).toBe('Sign in to KCD Bergen 2026')
+  })
+
+  it('sends as ITSELF once its domain is a platform sending domain', async () => {
+    vi.stubEnv('EMAIL_SENDING_DOMAINS', 'kcd.dev')
+    const sent = stubSender({ data: { id: 'email_1' } })
+
+    await sendEmailSignInLink(REQUEST)
+
+    expect(sent[0].from).toBe('KCD Bergen <hello@kcd.dev>')
+    expect(sent[0].replyTo).toBeUndefined()
+  })
+})
+
+describe('a rejected sign-in mail is observable', () => {
+  it('logs the conference and the sender at BOTH the client and the sign-in layer', async () => {
+    const error = vi.spyOn(console, 'error').mockImplementation(() => {})
+    stubSender({
+      error: { name: 'validation_error', message: 'domain is not verified' },
+    })
+
+    // The user-facing contract is unchanged: the caller learns nothing useful.
+    await expect(sendEmailSignInLink(REQUEST)).resolves.toBe(false)
+
+    const calls = error.mock.calls
+    expect(calls).toHaveLength(2)
+
+    // The client-level line names the tenant by its sender …
+    const chokePoint = calls.find((c) => c[0] === '[email] send failed')
+    expect(chokePoint?.[1]).toMatchObject({
+      from: 'KCD Bergen <noreply@platform.example>',
+      replyTo: 'hello@kcd.dev',
+      senderPolicy: 'platform-rewritten',
+    })
+
+    // … and the sign-in layer adds the conference and the org.
+    const signIn = calls.find(
+      (c) => c[0] === '[email-link] Resend rejected the sign-in email',
+    )
+    expect(signIn?.[1]).toMatchObject({
+      conference: 'KCD Bergen 2026',
+      orgId: 'org-kcd',
+      name: 'validation_error',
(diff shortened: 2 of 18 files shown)


Starefossen:
**The review finding was right, and fixing it surfaced two more holes.**

The **Effective senders** check now judges all three senders — `contactEmail`, `cfpEmail`, `sponsorEmail` — takes the worst verdict, and always names the offending address. Crucially the detail states that a healthy contact sender *proves nothing about the others*, so a partially-green reading cannot be misread as health.

The inventory is derived from the **send sites, not the schema** — all ~22 call sites re-walked — which matters here, because this PR already established that about half of them build `From:` inline rather than going through the helper. A schema-derived list would have looked right and been wrong.

**Two things found while fixing it:**

1. **The probe had the mirror-image blind spot.** `status.probeEmail` sent from `cfpEmail` unconditionally, so with a healthy CFP sender and an unverified sponsor one it returned green while sponsor mail was being refused. It now sends from the conference's WORST sender and reports `sentAs`, so a green result names what it actually proved.
2. **`resend.broadcasts.create` was never instrumented** — a second send API with its own `from`, used by the audience broadcast path. That was a real hole in this PR's "unbypassable client" claim, now closed.

**Verified independently rather than taken on trust.** Reverting the check to judge only `contactEmail`:

| state | result |
| --- | --- |
| all senders judged | 31/31 pass |
| `.slice(0, 1)` — contact only | **3 failed** — the REJECTED case, the rewrite case, and the per-field derivation case, each asserting the offender is NAMED |
| restored | 31/31 pass |

That is the property worth having: not that the check notices, but that it tells an operator WHICH sender is broken. A status page that says "something is wrong with email" sends someone hunting; one that says `REJECTED: CFP cfp@unverified.dev` does not.

Full suite 6729 pass · typecheck clean · eslint 0 errors / 216 warnings (byte-identical to main) · prettier clean · build succeeds.

⚠️ **The operational gate stands and is the owner's to clear:** `konf.app` is not verified in Resend, and both verified domains belong to tenants — tracked as RunKonf/platform#34.

Starefossen:
**Header injection fixed at the source — and the naive fix would have opened a second hole.**

`parseAddress()` sanitised the display name but returned the address raw in both branches, and line 231 fed that into `Reply-To`. Since the address comes from tenant-editable `contactEmail`/`cfpEmail`/`sponsorEmail`, an organizer storing CR/LF got bytes into a mail header — an injected `Bcc:` being the obvious payoff.

The exposure was wider than the one line: the ~10 send sites that build the header **inline** (rather than through `resolveConferenceFrom`, which already sanitised its field) handed `applySenderPolicy` a header with live CR/LF, and in the pass-through branches the raw `From:` reached Resend too.

## The part worth reading twice

Sanitisation **truncates at the first CR/LF** rather than deleting the breaks — and that distinction is itself a security property.

Deleting *splices* the payload onto the value. For an address, `a@evil.example\r\nb@verified.test` collapses to a single address whose domain reads as `verified.test` — so an unverified sender would classify as **verified and skip the rewrite**, defeating the very check this PR adds. The obvious fix would have created a bypass in the new code.

I verified that independently rather than taking it on trust:

| state | result |
| --- | --- |
| truncate at first break | 52/52 pass |
| **delete breaks instead** (the naive fix) | **3 failed** — including `truncation denies the attacker the resulting DOMAIN` |
| restored | 52/52 pass |

The test is named for the property, so a future refactor toward "just strip the newlines" fails with an explanation rather than a diff.

## Also

Every branch now rebuilds the header from parsed parts, so a caller-built header is never echoed — pinned separately (sabotage (f): 3 tests), which confirms that layer is not merely masked by the source fix. `resolveConferenceContact` reaches a `mailto:` href, badge-issuer JSON and email body copy — no SMTP header, and those encoders handle CR/LF — so not an injection path, but sanitised anyway so a future header consumer is not the next hole.

Full suite **6741 pass** · typecheck clean · eslint 0 errors / 216 warnings (unchanged from main) · prettier clean · build succeeds.

One behavioural note: `formatAddress(parseAddress(x))` drops surrounding quotes from a quoted display name. The helper already emitted unquoted names, so no output shape changes — it is the deliberate cost of never echoing a caller-built header.

Starefossen:
**Dedicated-client path verified independently. Enqueueing.**

The earlier fix put sanitisation inside `parseAddress`, which the POLICY path calls — but a tenant on its own Resend account skips the policy by design, so the ~10 send sites that build `from` inline handed a raw header straight to Resend. `parseAddress` was never the source; the source is the boundary where a message leaves for Resend.

It now sits in `instrumentResendClient`, unconditionally, on every client and every `from`-bearing method. Enumerating the SDK surface turned up two more entry points nobody had considered — `batch.send`/`batch.create` (an array of messages, each with its own `from`) and `broadcasts.update` (a broadcast created clean and updated poisoned). Both unused today, both wrapped, so adopting them later cannot reopen the hole.

**My verification, since the previous round's was too narrow:**

| state | result |
| --- | --- |
| guard intact | 56/56 pass |
| **client-boundary sanitiser made a pass-through**, policy untouched | **4 failed — all four DEDICATED-path tests**, 52 pass |
| restored | 56/56 pass |

That the shared-path injection tests all still pass under that sabotage is the point: it proves the new guard covers a path nothing else does, rather than being masked by defence in depth.

One process note worth recording. My **first** sabotage attempt was invalid — I removed the function body, which broke imports so 23 tests never executed, and "33 passed" looked like a clean run. I only caught it because the total did not match the restored count. A broken sabotage and an unnecessary guard produce the same-looking result, so the test COUNT matters as much as the pass/fail.

Full suite 6745 pass · typecheck clean · eslint 0 errors / 216 warnings (unchanged) · prettier clean · build succeeds.

⚠️ Operational gate unchanged: `konf.app` is not verified in Resend, so `EMAIL_FALLBACK_FROM` cannot yet point at the platform's own domain — RunKonf/platform#34.

PR #806, opened 2026-08-05
fix(security): close SSRF in proxy-image host allowlist

### LINKED ISSUE #732: SECURITY: proxy-image host check uses substring match — authenticated SSRF
Found by adversarial review of #731; out of scope for that PR, filed separately.

`src/app/api/proxy-image/route.ts:28` validates the target with `hostname.includes('sanity.io')` — a **substring** test, not a suffix test. So any host containing that string anywhere passes:

- `sanity.io.attacker.example`
- `attacker-sanity.io-cdn.example`
- `evil.com/?x=sanity.io` style variants depending on how the URL is parsed

The response is then served back **with `Access-Control-Allow-Origin: *` and a one-year CDN cache**, which turns a fetch primitive into a durable, publicly-readable, cross-origin-readable mirror of whatever the server could reach — including internal addresses if the runtime can route to them.

## Fix

Match on the parsed hostname with an exact suffix check against an allowlist (`=== 'cdn.sanity.io'`, or `endsWith('.sanity.io')` after confirming the parse), never `includes`. Reject non-https schemes, and consider dropping the wildcard CORS header and the long cache on a proxied response.

The repo has a precedent worth matching: the OAuth redirect callback had an equivalent `startsWith(baseUrl)` open-redirect that was fixed by comparing parsed origins rather than string prefixes.

Starefossen:
Closes #732.

## The bug

`src/app/api/proxy-image/route.ts` validated the fetch target with
`url.hostname.includes('sanity.io')` — a **substring** test. Every one of these
passed:

- `https://sanity.io.attacker.example/x.jpg`
- `https://attacker-sanity.io-cdn.example/x.jpg`
- `https://cdn.sanity.io.attacker.example/x.jpg`

The proxied body was then returned with `Access-Control-Allow-Origin: *` and
`s-maxage=31536000`, so a signed-in speaker could turn the server into a
durable, publicly readable, cross-origin mirror of anything the runtime could
route to — including link-local metadata endpoints and internal services.

## The fix

Exact equality against a one-entry allowlist:

```ts
const ALLOWED_IMAGE_HOST = 'cdn.sanity.io'

function isAllowedImageTarget(url: URL): boolean {
  return (
    url.protocol === 'https:' &&
    url.hostname === ALLOWED_IMAGE_HOST &&
    url.port === '' &&
    url.username === '' &&
    url.password === ''
  )
}
```

**Why exact, not `endsWith('.sanity.io')`:** the only thing this route is used
for is re-serving Sanity image assets so `html2canvas` can rasterize them
(`DownloadableImage`, `SpeakerSharingActions` are the sole callers). Sanity
serves every asset from `cdn.sanity.io` — the same single host `next.config.ts`
lists in `remotePatterns` and that `SANITY_CDN_PREFIX` in
`src/lib/sanity/client.ts` builds. A suffix check would additionally admit every
other `*.sanity.io` host for no caller's benefit, and is itself defeated by the
trailing-dot FQDN `cdn.sanity.io.`.

Note this does not narrow what works today: hosts such as
`avatars.githubusercontent.com` or `placehold.co` (which `speakerImageUrl` can
return) were already rejected by the old `sanity.io` check, so they are
unchanged — still 403, just for a correctly stated reason.

## Other SSRF levers

| Lever | Status |
| --- | --- |
| Substring host match | **Fixed** — exact `===` on the parsed hostname. |
| Trailing-dot FQDN `cdn.sanity.io.` | **Fixed** — different string, so `===` rejects it. Tested. |
| Userinfo `@` (`https://cdn.sanity.io@attacker.example/`) | **Fixed** — `URL` resolves the real host to `attacker.example`; additionally any non-empty `username`/`password` is rejected so credentials are never forwarded as an `Authorization` header. Tested. |
| Uppercase / mixed case host | **N/A by parsing** — `URL` lowercases; `HTTPS://CDN.SANITY.IO/...` still works. Regression-tested. |
| IDN / homograph (`cdn.sanitỵ.io`) | **N/A by parsing** — `URL` punycodes, so it never equals the ASCII host. Tested. |
| Non-https schemes | **Fixed** — `https:` only. `http://cdn.sanity.io` (downgrade) now 403s; `file:`/`data:` parse with an empty hostname and were already caught, now doubly so. Tested. |
| Non-default port | **Fixed** — `url.port === ''`. Tested. |
| **Redirect following** | **Fixed** — the fetch used the default `redirect: 'follow'`, so one 302 from a permitted host to `http://169.254.169.254/` would have defeated the allowlist entirely. Now `redirect: 'manual'`, and any 3xx (or `opaqueredirect` on a spec-strict runtime) returns 502. Tested both shapes. |
| Raw-string vs parsed target | **Fixed** — `fetch(url)` instead of `fetch(imageUrl)`, so the request target is provably the value that passed validation. |
| DNS rebinding on the allowed host | **N/A** — `cdn.sanity.io` is not attacker-controlled DNS. |
| `Access-Control-Allow-Origin: *` + 1-year public cache | **Kept, deliberately** — the body can now only come from one public CDN and is keyed by the full URL, so it carries nothing the caller's session made visible. Removing the header risks breaking `html2canvas` with `useCORS: true` for no security gain. Comment added at the call site stating the precondition. |
| SVG served from **our** origin | **Fixed (adjacent)** — `.svg` passes the extension and `image/*` content-type checks, so a navigation straight at a proxy URL would run CDN-hosted SVG script in our origin. Response now sets `X-Content-Type-Options: nosniff` and `Content-Security-Policy: default-src 'none'; sandbox`. Neither affects `` rendering. |
| Status-code passthrough as a probe oracle | **Accepted** — with the host pinned to one public CDN there is nothing to probe. |

Auth was already correct: the route 401s before any network access unless
`session.user` **and** `session.speaker` are present. Covered by a test.

## Sabotage proof

New suite: `src/app/api/proxy-image/route.test.ts`, **30 tests**.

1. Reverted the host check to the original `url.hostname.includes('sanity.io')`
   → **9 failed / 21 passed**. The failures are exactly the substring hosts, the
   trailing-dot FQDN, the `evil.sanity.io` suffix case, and the
   scheme/port/credential cases.
2. Restored, then removed `redirect: 'manual'` and the 3xx guard →
   **3 failed / 27 passed**.
3. Restored both → **30 passed**.

## Numbers

| Check | Result |
| --- | --- |
| `pnpm test` | **478 files / 6905 tests passed** (baseline 477 / 6875; +1 file, +30 tests — all new) |
| `pnpm typecheck` | clean, exit 0 |
| `npx prettier --check .` | "All matched files use Prettier code style!" |
| `npx eslint .` | 216 problems — **0 errors**, 216 warnings (unchanged from baseline) |

<h3>Greptile Summary</h3>

The PR closes the proxy-image SSRF path by restricting requests to parsed, credential-free HTTPS URLs on `cdn.sanity.io`, disabling redirect following, and hardening successful image responses.
- Replaces substring hostname matching with an exact URL-component allowlist.
- Fetches the validated `URL` object and rejects redirect responses.
- Adds response protections against MIME sniffing and active SVG navigation.
- Adds focused coverage for malicious targets, redirects, authentication, and response headers.

The PR appears safe to merge, with no actionable regressions identified in the changed proxy behavior.

The route now validates the complete parsed target before network access, prevents fetch redirects from escaping the allowlist, and safely hardens image responses; the tests cover the principal reachable attack paths.

| Filename | Overview |
|----------|----------|
| src/app/api/proxy-image/route.ts | Correctly closes the hostname-substring and redirect-based SSRF paths while preserving the intended Sanity image-proxy flow. |
| src/app/api/proxy-image/route.test.ts | Adds comprehensive regression coverage for target validation, redirects, authentication ordering, and response hardening. |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
  participant Speaker
  participant Proxy as proxy-image route
  participant Auth
  participant Sanity as cdn.sanity.io
  Speaker->>Proxy: "GET ?url=encoded target"
  Proxy->>Auth: Validate signed-in speaker
  Auth-->>Proxy: Authorized
  Proxy->>Proxy: Parse URL and verify HTTPS, exact host, port, credentials
  alt Target rejected
    Proxy-->>Speaker: 403 Invalid image source
  else Target allowed
    Proxy->>Sanity: fetch(validated URL, redirect: manual)
    alt Redirect response
      Sanity-->>Proxy: 3xx / opaqueredirect
      Proxy-->>Speaker: 502 Image source redirected
    else Valid image response
      Sanity-->>Proxy: image bytes
      Proxy-->>Speaker: Image with CORS, cache, nosniff, and CSP headers
    end
  end
```

<sub>Reviews (1): Last reviewed commit: ["fix(security): close SSRF in proxy-image..."](https://github.com/cloudnativebergen/website/commit/4942836bb2409f4a4985050a268cc927e9885570) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=50354288)</sub>

### DIFF
--- a/src/app/api/proxy-image/route.test.ts
+++ b/src/app/api/proxy-image/route.test.ts
@@ -1,0 +1,227 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest'
+import type { NextRequest } from 'next/server'
+
+const authMock = vi.fn()
+vi.mock('@/lib/auth', () => ({
+  auth: () => authMock(),
+}))
+
+vi.mock('next/cache', () => ({
+  unstable_noStore: () => {},
+}))
+
+import { GET } from './route'
+
+/**
+ * The route only ever reads `request.url`, so a bare stand-in keeps these tests
+ * free of Next's request plumbing.
+ */
+const requestFor = (target: string) =>
+  ({
+    url: `https://conference.example/api/proxy-image?url=${encodeURIComponent(target)}`,
+  }) as unknown as NextRequest
+
+const imageResponse = () =>
+  new Response(new Uint8Array([1, 2, 3]), {
+    status: 200,
+    headers: { 'content-type': 'image/png' },
+  })
+
+let fetchMock: ReturnType<typeof vi.fn>
+
+beforeEach(() => {
+  vi.clearAllMocks()
+  authMock.mockResolvedValue({
+    user: { email: 'a@b.c' },
+    speaker: { _id: 's' },
+  })
+  fetchMock = vi.fn().mockResolvedValue(imageResponse())
+  vi.stubGlobal('fetch', fetchMock)
+  vi.spyOn(console, 'error').mockImplementation(() => {})
+})
+
+afterEach(() => {
+  vi.unstubAllGlobals()
+  vi.restoreAllMocks()
+})
+
+describe('/api/proxy-image — host allowlist', () => {
+  it('proxies a genuine cdn.sanity.io image', async () => {
+    const res = await GET(
+      requestFor('https://cdn.sanity.io/images/p/d/abc123-400x400.jpg'),
+    )
+
+    expect(res.status).toBe(200)
+    expect(res.headers.get('content-type')).toBe('image/png')
+    expect(fetchMock).toHaveBeenCalledTimes(1)
+    expect(String(fetchMock.mock.calls[0][0])).toBe(
+      'https://cdn.sanity.io/images/p/d/abc123-400x400.jpg',
+    )
+  })
+
+  it('proxies a Sanity URL that carries transform query params', async () => {
+    const res = await GET(
+      requestFor('https://cdn.sanity.io/images/p/d/abc123-400x400?w=400&q=85'),
+    )
+
+    expect(res.status).toBe(200)
+  })
+
+  it('accepts an uppercase host, since URL parsing lowercases it', async () => {
+    const res = await GET(
+      requestFor('HTTPS://CDN.SANITY.IO/images/p/d/abc123-400x400.jpg'),
+    )
+
+    expect(res.status).toBe(200)
+  })
+
+  /**
+   * The reported bug (#732): `hostname.includes('sanity.io')` is a SUBSTRING
+   * test, so every host below passed it and turned this route into an
+   * authenticated SSRF primitive with `Access-Control-Allow-Origin: *` and a
+   * one-year CDN cache.
+   */
+  it.each([
+    // Contains the allowed domain but does not end with it.
+    'https://sanity.io.attacker.example/x.jpg',
+    'https://attacker-sanity.io-cdn.example/x.jpg',
+    'https://cdn.sanity.io.attacker.example/x.jpg',
+    // Ends with the domain but is not the allowed host.
+    'https://evil.sanity.io/x.jpg',
+    'https://notsanity.io/x.jpg',
+    // Trailing-dot FQDN: resolves the same, but is a different string — the
+    // classic defeat for an `endsWith` allowlist.
+    'https://cdn.sanity.io./x.jpg',
+    // Userinfo trick: everything before `@` is credentials, not the host.
+    'https://cdn.sanity.io@attacker.example/x.jpg',
+    'https://cdn.sanity.io%2f@attacker.example/x.jpg',
+    // Homograph / IDN: parsed to punycode, so it never equals the ASCII host.
+    'https://cdn.sanitỵ.io/x.jpg',
+    // The domain smuggled somewhere other than the host.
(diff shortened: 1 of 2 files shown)


Starefossen:
Both review threads resolved, neither blocking:

1. **The "redacted string" test finding is a false positive from the reviewer's own display layer.** The committed line is `https://user:pass@cdn.sanity.io/x.jpg` — a legitimate userinfo-bypass test case. The reviewer redacted the `user:pass@` to `******` when echoing it back, then read its own redaction as a parse failure. The case resolves its hostname to `cdn.sanity.io`, is rejected by the non-empty-`username` check, and asserts rejection. Independent QA confirmed it.
2. **clearTimeout-after-await** is a real but minor pre-existing resource nit (a rejected fetch leaks the 10s timer). Filed with the other residual proxy-route items rather than expanding this security PR.

PR #824, opened 2026-08-05
fix(cfp): stop a freshly provisioned tenant's public CFP page from 500ing

Starefossen:
## The bug

`src/lib/onboarding/create.ts:167-171` deliberately provisions a new tenant with **no `formats` and no `topics`**, on the stated reasoning that "the admin surfaces are empty-safe" and "the activation checklist walks the organizer through CFP configuration". Both halves fail.

1. **The public surfaces are not empty-safe.** The conference GROQ is a bare `...` spread, so an absent field comes back `undefined`, not `[]` — while `Conference` declares `formats: Format[]`. `src/app/(main)/cfp/page.tsx:53` did `conference.formats.filter(...)` unguarded. There is **no `error.tsx` anywhere under `src/app`**, so that is a bare Next 500 on the first CFP link a new organizer shares with speakers. The tell: every *admin* call site already guards with `?? []`; the public ones did not.
2. **The checklist had no `formats` row**, so the remedy the comment defers to did not exist — even though a format is genuinely REQUIRED to submit (`src/lib/proposal/validation.ts`).

## Unguarded sites found

Swept `.formats` / `.topics` across `src/app/(main)`, `src/app/(cfp)`, `src/components/cfp`, `src/components/proposal` and the rest of the tree. **Three** genuinely unguarded reads, all public — the same three reported, no more:

| Site | Read |
|---|---|
| `src/app/(main)/cfp/page.tsx:53,56` | `conference.formats.filter(...)` x2 |
| `src/components/cfp/ProposalGuidanceSidebar.tsx:75` | `conference.formats.map(...)` |
| `src/components/proposal/ProposalDetailsForm.tsx:272` | `allowedFormats.includes(...)`, fed `conference.formats` by `(cfp)/cfp/proposal/page.tsx:181` and `[id]/page.tsx:201` |

Everything else already guards (`conference.topics || []`, `conference.formats ?? []`, `Array.isArray(...)`, `talk.topics && ...`). `src/app/(main)/program/page.tsx:122`, the admin settings page, `ProposalManagementModal`, `ProposalsFilter`, `useProposalFilters` etc. were all already safe.

## Where the fix lives, and why

**At the data boundary**, not as `?? []` at every use. `normalizeConference` (`src/lib/conference/normalize.ts`) is applied on every exit path of `getConferenceForDomain`, and coerces the fields the type declares NON-optional but the document routinely lacks: `formats`, `topics`, `domains`, `organizers`.

Rationale: there are dozens of reads across pages, components and stories, and every one is a place a future change can forget the guard. Sprinkling `?? []` leaves the type lying and the next unguarded read one commit away. Normalising where the document enters the app makes the non-optional array types *true* for every consumer instead of merely usually-true. (`domains` is in the list because provisioning also omits it when the operator supplies none — same crash class, one line away.)

Two exceptions keep a local guard, deliberately: `ProposalGuidanceSidebar` and `ProposalDetailsForm` take a `Conference` / `Format[]` as a **prop** and are rendered from Storybook and from admin surfaces that build their own objects, so they never see the boundary. `allowedFormats` becomes optional-with-default, matching the already-optional contract on the admin `ProposalsFilter` / `ProposalsList`.

Two other fetchers in the same module — `getConferencesForWeeklyUpdate` and `getConferenceByCheckinEventId` — are deliberately **outside** the boundary: their only consumers (the weekly-update Slack cron, the ticket-sold webhook) never read the four arrays. That is recorded in the normaliser's doc comment so the next reader does not have to re-derive it.

## Activation checklist

New required `formats` row — "At least one session format" — mirroring the `topics` row (same `#team-content` anchor, which is where the `FormatsEditor` lives). Without it the checklist reported "ready to launch" for a CFP that could accept nothing.

## CFP open with no formats

**Decided in this PR, not deferred** — but *without* widening `isCfpOpen`.

`isCfpOpen` answers a purely date-based question and is consumed by the homepage lifecycle stage, the admin phase summary, and the edit/unsubmit gates on proposals that already exist. Folding a formats requirement into it would silently change all of them. Instead there is a separate `hasSubmittableFormats`, paired with `isCfpOpen` only on surfaces that invite a **new** submission:

- `/cfp` — the "Submit your proposal" CTA is replaced by an honest `SubmissionsNotOpenNotice` ("Submissions are not open yet... the session formats have not been announced", with the CFP/contact address when set). Fact rows with nothing in them are omitted rather than rendered as an empty heading.
- `/cfp/proposal` — a "Submissions Not Open Yet" card instead of a form whose Format dropdown has no options. The pre-existing "CFP Closed" branch is untouched.
- **Both server routes that reach `submitted`:**
  - `proposal.create`, scoped to non-drafts. A draft is the incomplete-work path (it already skips strict validation), so an API/CLI caller can still prepare one before formats are announced.
  - `proposal.action`, on the draft -> submitted transition — the path `ProposalForm` uses for an existing draft, from an edit page with no gate of its own. Without this the draft carve-out above *was* the bypass: prepare a draft, then submit it here, landing a proposal carrying `ProposalDraftSchema`'s default `lightning_10` that the conference never offered. Unlike the 3-proposal cap beside it, this gate has **no organizer carve-out**: the cap is a per-speaker fairness rule an organizer may override, whereas a submitted proposal on a formats-less conference is wrong whoever creates it.

  `update` cannot flip status (no `status` in its schema) and there is no REST proposal endpoint, so those two are the whole surface. Editing and unsubmitting an existing proposal stay on the plain `isCfpOpen` window.

## Also fixed (found by looking at it)

The sidebar's **"Important Dates"** card rendered a heading over an empty list on a fresh tenant — every date it shows is optional and provisioning sets none. Same emptiness class, so it is hidden too. Verified visually with `pnpm shoot` (new `FreshlyProvisionedTenant` story); the configured `Default` story is unchanged.

## Not done here

**No `error.tsx` boundary added** — filed as #825 instead. It is a real finding (any unguarded error in this app is currently a bare 500) but it is an app-wide UX surface that wants its own design pass, and adding it would have masked the very crash this PR is proving is gone.

## Tests

5 new files; **96 tests across the PR's seven test files, up from 59 on the two it modifies — +37**.

The one that matters: `__tests__/app/cfp/fresh-tenant.test.tsx` builds the conference document with the **real `buildOnboardingDocuments`** — not a fixture that looks fresh — pushes it through the real `getConferenceForDomain`, and walks the public CFP page. It also asserts the premise (provisioning really does omit `formats`, `topics`, and `domains` when none are supplied), so it fails loudly if provisioning ever changes.

**Sabotage-proven.** Each guard removed by exact string, matching test confirmed failing, restored, confirmed green:

| Guard | Sabotaged result |
|---|---|
| G1 boundary normaliser (`conference/sanity.ts`) | 5 failed / 3 passed |
| G2 sidebar Accepted-Formats guard | 3 failed / 3 passed |
| G3 `allowedFormats = []` default | 1 failed / 5 passed |
| G4 Important-Dates emptiness guard | 1 failed / 5 passed |
| B1 activation `formats` row | 5 failed / 18 passed |
| B2a public CFP submit-CTA gate | 1 failed / 7 passed |
| B2b submit-page gate | 1 failed / 1 passed |
| B2c server `proposal.create` gate | 1 failed / 44 passed |
| B2d server `proposal.action` draft -> submitted gate | 2 failed / 43 passed |
| **restored** | **84 passed (84)** — the five files the sabotage targets, not the PR total |

## Numbers

| Check | Baseline (`origin/main` @ b616f2ab) | This branch |
|---|---|---|
| `pnpm test` | 487 files / 7008 tests, **1 failed** (`migrations/046.../plan.test.ts` canvas-rasterize timeout; passes in isolation — pre-existing flake) | **492 files / 7045 tests, 0 failed** |
| `npx eslint .` | 0 errors, 216 warnings | 0 errors, 216 warnings |
| `pnpm knip` | 1 pre-existing config hint | same hint, nothing new |

### DIFF
--- a/__tests__/api/trpc/proposal.test.ts
+++ b/__tests__/api/trpc/proposal.test.ts
@@ -680,6 +680,75 @@ describe('proposal router', () => {
       ).rejects.toMatchObject({ code: 'BAD_REQUEST' })
     })
 
+    it('should refuse submitting a draft when the conference offers no formats', async () => {
+      // `action` is the OTHER route to `submitted` — it is what ProposalForm
+      // uses for an existing draft. Gating only `proposal.create` would leave
+      // the trap open: prepare a draft (still allowed), then submit it here
+      // onto a conference offering no formats, carrying the schema's default
+      // format that the conference never offered.
+      vi.mocked(isCfpOpen).mockReturnValue(true)
+      vi.mocked(hasSubmittableFormats).mockReturnValue(false)
+      vi.mocked(getProposalSanity).mockResolvedValue({
+        proposal: { ...mockProposal, status: Status.draft } as any,
+        proposalError: null,
+      })
+      vi.mocked(getProposals).mockResolvedValue({
+        proposals: [],
+        proposalsError: null,
+      })
+      vi.mocked(updateProposalStatus).mockClear()
+
+      const caller = createAuthenticatedCaller(regularSpeaker._id)
+      await expect(
+        caller.proposal.action({ id: 'proposal-1', action: Action.submit }),
+      ).rejects.toMatchObject({
+        code: 'FORBIDDEN',
+        message: expect.stringContaining('session formats'),
+      })
+      expect(updateProposalStatus).not.toHaveBeenCalled()
+    })
+
+    it('should refuse an ORGANIZER submitting a draft with no formats either', async () => {
+      // No organizer carve-out: a submitted proposal on a formats-less
+      // conference is wrong whoever creates it (unlike the 3-proposal cap,
+      // which is a per-speaker fairness rule organizers may override).
+      vi.mocked(isCfpOpen).mockReturnValue(true)
+      vi.mocked(hasSubmittableFormats).mockReturnValue(false)
+      vi.mocked(getProposalSanity).mockResolvedValue({
+        proposal: { ...mockProposal, status: Status.draft } as any,
+        proposalError: null,
+      })
+
+      const caller = createAdminCaller()
+      await expect(
+        caller.proposal.action({ id: 'proposal-1', action: Action.submit }),
+      ).rejects.toMatchObject({ code: 'FORBIDDEN' })
+    })
+
+    it('should allow submitting a draft once formats are configured', async () => {
+      vi.mocked(isCfpOpen).mockReturnValue(true)
+      vi.mocked(hasSubmittableFormats).mockReturnValue(true)
+      vi.mocked(getProposalSanity).mockResolvedValue({
+        proposal: { ...mockProposal, status: Status.draft } as any,
+        proposalError: null,
+      })
+      vi.mocked(getProposals).mockResolvedValue({
+        proposals: [],
+        proposalsError: null,
+      })
+      vi.mocked(updateProposalStatus).mockResolvedValue({
+        proposal: { ...mockProposal, status: Status.submitted } as any,
+        err: null,
+      })
+
+      const caller = createAuthenticatedCaller(regularSpeaker._id)
+      const result = await caller.proposal.action({
+        id: 'proposal-1',
+        action: Action.submit,
+      })
+      expect(result.proposalStatus).toBe(Status.submitted)
+    })
+
     it('should enforce proposal cap when submitting a draft', async () => {
       vi.mocked(isCfpOpen).mockReturnValue(true)
       const draftProposal = { ...mockProposal, status: Status.draft }
--- a/__tests__/api/trpc/proposal.test.ts
+++ b/__tests__/api/trpc/proposal.test.ts
@@ -444,6 +444,28 @@ describe('proposal router', () => {
       ).rejects.toMatchObject({ code: 'FORBIDDEN' })
       expect(createProposal).not.toHaveBeenCalled()
     })
+
+    it('still allows a DRAFT when the conference offers no formats', async () => {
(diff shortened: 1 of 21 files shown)


Starefossen:
Both open threads were addressed in later commits; resolving with the trail, because the two findings chain into each other.

**Greptile — the gate blocked draft creation.** Correct, and fixed: the gate is now scoped to **submission**, not draft creation. A draft is explicitly the incomplete-work path, and an API/CLI caller must stay able to prepare one before organizers announce their formats.

That fix is also what made the next finding reachable — adversarial QA showed `proposal.action` is a **second** route to `submitted`, so a draft prepared under the carve-out could then be submitted onto a formats-less conference, carrying `ProposalDraftSchema`'s default `lightning_10`, a format the conference never offered. The gate now sits on the `draft → submitted` transition itself, deliberately **without** an organizer carve-out: the neighbouring 3-proposal cap is a per-speaker fairness rule an organizer may override, whereas a submitted proposal on a conference offering no formats is wrong whoever creates it. The router comment that claimed "this refuses a direct call" was false for that route and now names both.

**Copilot — the misleading `Pick<Conference,'formats'>` parameter type.** Also correct and fixed: the implementation tolerates an absent/invalid `formats`, so the type now says so rather than implying the field is always present.

Sabotage coverage across the PR is nine guards; the new `proposal.action` case fails 2 tests when removed.

PR #833, opened 2026-08-05
feat(onboarding): provision new conferences with starter session formats

Starefossen:
## Decision

Provisioning deliberately wrote **no** `formats`, with a comment saying the activation checklist would walk the organizer through it. That is what caused the day-one CFP crash fixed in #824. Owner decision: ship starter formats so a fresh tenant's CFP does not depend on an organizer finding a checklist row.

**Topics stay empty** — a topic list is far more conference-specific than a session length, so any seed would be one conference's subject matter imposed on every tenant.

## The starter set — and why

`lightning_10`, `presentation_25`, `presentation_45`. Ids come from the fixed `Format` enum (`src/lib/proposal/types.ts`); nothing invented.

| Format | Why |
| --- | --- |
| `lightning_10` | The short-slot staple, **and** the format `ProposalDraftSchema` falls back to when a draft carries none. Seeding it means that schema default is always a format the conference actually offers — the exact mismatch `proposal.action`'s gate warns about. |
| `presentation_25` | Fits a 30-minute grid slot with changeover. |
| `presentation_45` | Fits a 60-minute grid slot with changeover. |

**Deliberately absent:** workshops (`workshop_120`/`workshop_240`) commit a conference to rooms, instructors and a separate track — the public CFP page renders an entire "Hands-on Workshops" section promising them — so they are opt-in, never a default. `presentation_20`/`presentation_40` are omitted as near-duplicates; a starter list offering five talk lengths is a menu to prune, not a default to accept.

## The finding: starter formats alone were NOT enough

Verifying end to end rather than assuming turned up the other half of the #824 trap. **A submitted proposal requires at least one TOPIC too** (`validateProposalForm`, `ProposalInputSchema.topics.min(1)`), and the submit form's topic picker is populated purely from `conference.topics`.

So starter formats on their own would have made day one *worse*: the public CFP page would have shown "Submit your proposal", handed the speaker a form with a working format picker and an **empty dropdown over a required field**, and rejected the submission — replacing an honest "Submissions are not open yet" with a dead end. Same bug class as #824, one field over.

Fix: `canAcceptProposals(conference)` = `hasSubmittableFormats && hasSubmittableTopics`, used by the two surfaces that invite a NEW submission (`/cfp` and `/cfp/proposal`). Both already fetch with `{ topics: true }`.

`src/server/routers/proposal.ts` deliberately keeps the narrower `hasSubmittableFormats` gate: it calls `getConferenceForCurrentDomain()` **without** projecting topics, so a topic-aware gate there would read `[]` from the boundary normaliser and fail CLOSED on every well-configured conference. What refuses a topic-less proposal there is the strict **content** validation on both submit paths — see the next section, which is where that turned out not to be true yet.

Owner decision on this finding: **the remaining topic step stays a checklist step.** Net effect on day one: one of the two blockers is gone, and adding topics now needs no format work.

## Second commit: `action` promoted drafts with no content validation at all

Adversarial QA probed the carve-out above and found the comment was making a claim the code did not back. `create` strict-parses a proposal before it can reach `submitted`; **`action`'s draft→submitted transition parsed nothing** — only the formats gate and the proposal cap. QA promoted a draft with `topics: []` and `tos: false` to `submitted` on a topics-less fresh tenant. Accepted, no error.

The hole is **pre-existing on any conference**: `ProposalDraftSchema` is `.partial()`, so `create({ status: draft })` accepts anything and `action(submit)` then promoted it — bypassing every content rule `create` enforces. The `update` route has the strict-validation-on-merged-content pattern; `action` never got it. What this PR changes is that fresh tenants were previously shielded by the formats gate, and seeding formats removes that accidental shield.

Fixed rather than merely re-documented, because the UI cannot reach it but **direct API and tRPC callers can** — a real population given the agent-facing CLI direction (`konfctl`, MCP server in the works).

Applies the `update` route's pattern with nothing to merge: the stored document is parsed. Two fields are folded from read shape back to what the schema describes — `speakers` dropped, and **`topics` folded from dereferenced topic documents back to references**. The projection returns `topics[]-> { _id, title, color }`, so parsing the document naively would have rejected every legitimate submission on the topic field; there is a named regression test for exactly that. No organizer carve-out, matching the formats gate. Scoped to draft→submitted, so withdraw/unsubmit on an already-imperfect proposal still work.

The `canAcceptProposals` docstring is corrected to say what is now true.

### Third asymmetry found, NOT fixed — needs an owner call

While auditing "what else guards `create` but not `action`", I probed the CFP window and confirmed it live:

- `create` refuses **any** call — even a draft — when `isCfpOpen` is false.
- `action`'s draft→submitted has **no window check at all**. A speaker sitting on an old draft can promote it to `submitted` a month after the CFP closed. (`action` *does* block `unsubmit` after close, so the omission looks like an oversight rather than a design.)

Left alone deliberately: it is unrelated to starter formats, it is pre-existing, and closing it changes behaviour for organizers too (who may legitimately promote drafts after close on a speaker's behalf). Happy to fix in a follow-up — it wants a decision, not a guess.

## Checklist row

Row starts ticked — correctly, because the CFP genuinely has formats — with a `note` that keeps the tick honest:

> ~~Session formats~~
> We started you off with Lightning Talk (10 min), Presentation (25 min) and Presentation (45 min) — edit them to match your programme.

`ActivationRow.note` is new: an advisory shown even when a row is done, for a requirement satisfied by something the organizer did not choose. Rare by design. Format titles come from the same map the editor and CFP page render, so the checklist can never name a format differently from the surface it links to.

Verified visually in Storybook (light + dark) via `pnpm shoot`.

## Judgement calls

- **Are starter formats marked as defaults anywhere in admin? No** — agreeing with the stated instinct. They are ordinary formats once written; a "still the defaults" flag would be state to maintain forever for no behaviour. The checklist note instead **derives** "untouched starter set" from the list itself (`isUntouchedStarterFormatSet`), so any edit retires it with nothing to clean up.
- **The known false positive is accepted and now documented as such.** A conference that never saw the starter set — created before it existed, or whose organizer independently landed on precisely these three — reads as untouched, so the note claims a provenance untrue of it. Cost: one line of slightly wrong attribution on an already-satisfied row, self-healing on the next edit. The alternative is a stored flag, which is exactly what this design avoids. The docstring no longer claims "the copy is still true".
- **Fixed alongside:** the set-membership compare accepted a four-entry list containing a duplicated starter member. Now length-checked against the array.
- **The checklist row stays `required` and stays ticked on a fresh tenant.** Marking it outstanding would be a false launch blocker — the formats really are there.
- **Copy on `SubmissionsNotOpenNotice` no longer names formats specifically**, now that either half can be the missing one. A speaker cannot act on "topics are missing"; the organizer reads the checklist, not this notice.

## Existing conferences

**Unaffected.** This changes provisioning only — no migration, no backfill, no script. Existing conferences keep whatever formats they have.

The one behaviour change that could reach them is the `canAcceptProposals` gate. In practice it cannot: `conference.topics` is `Rule.required().min(1)` in `sanity/schemaTypes/conference.ts`, so any conference an organizer has edited already has topics. And even if one somehow had none, the gate would only withhold a CTA from a conference that could not have accepted the submission anyway.

The `action` validation guard does reach existing conferences — deliberately. It refuses only promotions that `create` would already have refused, and the UI path (`update` then `action`) always saves complete content first.

## Sabotage proof

**Starter formats.** Removing the single line `formats: [...STARTER_SESSION_FORMATS],` from `create.ts`:

- **14 tests fail across 3 files** (`src/lib/onboarding/create.test.ts`, `__tests__/app/cfp/fresh-tenant.test.tsx`, `__tests__/app/cfp/fresh-tenant-submit.test.tsx`) — including "seeds the exact starter set", "is provisioned WITH the starter formats", "scopes the format picker to exactly the starter formats", and both `proposal.create` / `proposal.action` submissions.
- Restored → green again.

**The new validation guard.** Removing the `ProposalInputSchema.safeParse` block from `action`:

- **5 tests fail across 2 files** — "refuses a draft carrying no topics", "refuses a draft with the terms unaccepted", "refuses a draft with an empty description", "gives ORGANIZERS no carve-out either", and the QA repro "does NOT let the two-step path smuggle a topic-less proposal through".
- Restored → green again.

Tests build the fixture with the **real** `buildOnboardingDocuments` (reusing #824's precedent), assert the starter list **literally** rather than recomputing it from the constant, and run the **real** `hasSubmittableFormats` / `canAcceptProposals` (`__tests__/api/trpc/proposal.test.ts` mocks that predicate, so it can never tell you whether a real provisioned conference passes it).

## Numbers

Rebased onto `origin/main` @ `1199fc01` (#830). Baseline re-derived on that commit in a clean worktree.

| Check | Baseline `origin/main` | Branch |
| --- | --- | --- |
| `pnpm test` | 501 files / 7169 tests pass | **502 files / 7208 tests pass** (+1 file, +39) |
| `npx eslint .` | 209 problems (0 errors, 209 warnings) | 209 problems (0 errors, 209 warnings) |
| `pnpm knip` | exit 0 | exit 0 |

### DIFF
--- a/__tests__/api/trpc/proposal.test.ts
+++ b/__tests__/api/trpc/proposal.test.ts
@@ -749,6 +749,128 @@ describe('proposal router', () => {
       expect(result.proposalStatus).toBe(Status.submitted)
     })
 
+    describe('strict content validation on draft \u2192 submitted', () => {
+      // `create` strict-parses the payload before letting a proposal reach
+      // `submitted`; `action` \u2014 the OTHER route to `submitted` \u2014 did not parse
+      // anything. Since `ProposalDraftSchema` is `.partial()`, the two-step path
+      // `create({ status: draft })` \u2192 `action(submit)` promoted content that
+      // `create({ status: submitted })` would have refused outright. The UI
+      // cannot reach it (`ProposalForm` saves via `update` first), a direct
+      // API/tRPC caller can.
+      function draft(overrides: Record<string, unknown> = {}) {
+        vi.mocked(isCfpOpen).mockReturnValue(true)
+        vi.mocked(hasSubmittableFormats).mockReturnValue(true)
+        vi.mocked(getProposalSanity).mockResolvedValue({
+          proposal: {
+            ...mockProposal,
+            status: Status.draft,
+            ...overrides,
+          } as any,
+          proposalError: null,
+        })
+        vi.mocked(getProposals).mockResolvedValue({
+          proposals: [],
+          proposalsError: null,
+        })
+        vi.mocked(updateProposalStatus).mockClear()
+      }
+
+      it('refuses a draft carrying no topics', async () => {
+        draft({ topics: [] })
+
+        const caller = createAuthenticatedCaller(regularSpeaker._id)
+        await expect(
+          caller.proposal.action({ id: 'proposal-1', action: Action.submit }),
+        ).rejects.toMatchObject({
+          code: 'BAD_REQUEST',
+          message: expect.stringContaining('At least one topic is required'),
+        })
+        expect(updateProposalStatus).not.toHaveBeenCalled()
+      })
+
+      it('refuses a draft with the terms unaccepted', async () => {
+        draft({ tos: false })
+
+        const caller = createAuthenticatedCaller(regularSpeaker._id)
+        await expect(
+          caller.proposal.action({ id: 'proposal-1', action: Action.submit }),
+        ).rejects.toMatchObject({ code: 'BAD_REQUEST' })
+        expect(updateProposalStatus).not.toHaveBeenCalled()
+      })
+
+      it('refuses a draft with an empty description', async () => {
+        draft({ description: [] })
+
+        const caller = createAuthenticatedCaller(regularSpeaker._id)
+        await expect(
+          caller.proposal.action({ id: 'proposal-1', action: Action.submit }),
+        ).rejects.toMatchObject({ code: 'BAD_REQUEST' })
+      })
+
+      it('gives ORGANIZERS no carve-out either', async () => {
+        // Same reasoning as the format gate: an invalid submitted proposal is
+        // wrong whoever promotes it. (Unlike the 3-proposal cap, which is a
+        // per-speaker fairness rule organizers may override.)
+        draft({ topics: [] })
+
+        const caller = createAdminCaller()
+        await expect(
+          caller.proposal.action({ id: 'proposal-1', action: Action.submit }),
+        ).rejects.toMatchObject({ code: 'BAD_REQUEST' })
+      })
+
+      it('accepts DEREFERENCED topics \u2014 the shape the projection returns', async () => {
+        // REGRESSION GUARD. `getProposal` projects `topics[]-> { _id, title }`,
+        // not `{ _type: 'reference', _ref }`. Parsing the stored document
+        // naively against `ProposalInputSchema` would reject every legitimate
+        // submission on the topic field, so the gate folds them back first.
+        draft({
+          topics: [
+            { _id: 'topic-1', title: 'Platform engineering', color: '#fff' },
+          ],
+        })
+        vi.mocked(updateProposalStatus).mockResolvedValue({
+          proposal: { ...mockProposal, status: Status.submitted } as any,
+          err: null,
+        })
+
+        const caller = createAuthenticatedCaller(regularSpeaker._id)
+        const result = await caller.proposal.action({
+          id: 'proposal-1',
+          action: Action.submit,
+        })
+        expect(result.proposalStatus).toBe(Status.submitted)
+      })
+
+      it('does NOT validate content on other transitions', async () => {
+        // Withdrawing or unsubmitting an already-imperfect proposal must stay
+        // possible \u2014 the gate is scoped to the draft \u2192 submitted promotion.
+        vi.mocked(isCfpOpen).mockReturnValue(true)
+        vi.mocked(isWithdrawalCutoffActive).mockReturnValue(false)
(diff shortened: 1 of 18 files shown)


Starefossen:
**Owner decision (Hans, 2026-08-06) on the topic blocker you surfaced: leave it — one checklist step is fine.**

Your finding stands and was the right call to raise: seeding formats alone would have made day one *worse*, replacing an honest "submissions are not open yet" with a form that has a working format picker and an empty dropdown over a required field. The `canAcceptProposals` gate is the correct resolution.

Rejected alternatives, for the record: making topics optional for submission (they drive review filtering and programme building — weakening the rule everywhere to fix a day-one case is the wrong trade), seeding generic starter topics (topics are genuinely conference-specific in a way session lengths are not), and a conditional "required only when the conference defines topics" rule (more validation machinery to reason about than the problem warrants).

So the shipped state is intended: **formats seeded, topics added by the organizer via the activation checklist, and the public CFP honestly showing "not open yet" until then.** No follow-up needed on the mandatory-topic rule — please make sure the provisioning comment and the checklist copy read as a deliberate choice rather than an unfinished one.

Starefossen:
Fixed in the latest commit — the length check now compares against the array rather than the deduplicated set, so a four-entry list containing a repeated starter member is correctly treated as edited. Test added.

The docstring also stopped overstating: the remaining false positive (a legacy conference whose organizer independently chose exactly these three formats reads as untouched and gets a provenance line that is literally wrong) is now an explicit **KNOWN FALSE POSITIVE, accepted** paragraph rather than the previous "the copy is still true" stretch. Cost is one slightly wrong line on an already-satisfied checklist row, self-healing on any edit; the alternative is a stored "is starter" flag, which is exactly the state this design avoids.

PR #828, opened 2026-08-05
feat(admin): gate ticketing and badges instead of dead-ending tenants

Starefossen:
## Why

An audit of a brand-new tenant's day one found the two worst admin surfaces, both about to be shown to sales prospects in a demo organization (`RunKonf/platform#47`):

1. **Badge admin had no gate at all.** `issueBadgeForSpeaker` refuses every non-platform org (the Phase 0 tripwire, `RunKonf/platform#46`), but `/admin/speakers/badge` rendered in full for every tenant and `BadgeManagementClient` surfaced the raw refusal **per speaker** — a message naming our internal issue tracker.
2. **All five provider-backed ticket routes rendered a red `ErrorDisplay "Checkin.no Configuration Error"`** when the binding was absent: an error frame for "not set up yet", with no link to settings. Plus two aggravations — the guard hardcoded the **Checkin** fields even though `resolveTicketingProvider` fully supports **Tito** (so a Tito-bound conference could never open *any* ticket page), and after #820 an org that fills in the ids but has no credentials still failed: raw `"Missing checkin configuration"` on the overview, a false *"No tickets have been sold"* on orders, and a silently empty list on types.

The structural cause: the nav/⌘K filtering machinery existed but could only gate one feature. `src/lib/features/registry.ts` knew `graphql-api | dedicated-email | slack-mirror | workshops`, and the admin layout computed `enabledFeatures` as literally `['workshops']` or `[]`.

## The structural enabler

- **`ticketing` and `badges` added to `FEATURE_IDS`**, both `readiness: 'internal'` with **no `minPlan`** — matching today's reality (both are platform-org-gated in code already) and deliberately **not** guessing a plan tier. A test fails the moment someone adds one.
- **`src/lib/features/platform-default.ts`** factors out the resolution `workshops` worked out first (registry decision → overrides win both ways → platform-org default, fail-closed at every step). `workshops.ts` now delegates to it, byte-identical behaviour.
- **`ticketing.ts`** layers one extra grant: an org with its **own** provider credentials in the per-org secret store. The gate must never be stricter than `resolveTicketingCredentials`, or it would hide a surface that works.
- **`enabled.ts`** composes the registry resolution with each gate, and the layout now calls it instead of the hardcoded array. Tagging a destination is now all it takes to gate it.
[...]

PR #834, opened 2026-08-05
feat(admin): sell ticketing at the entry paid tier, make a deny a hard kill switch

Starefossen:
Applies two owner decisions (2026-08-06) on top of the gating that landed in #828.

## 1. Ticketing moves to the entry PAID tier; badges stay internal

`ticketing` is now `readiness: 'ga'` with `minPlan: 'pro'`. The plan ladder in this codebase is **community (free/comped) < pro < enterprise** (`ORGANIZATION_PLANS`), so the lowest tier a customer can actually *buy* is `pro` — that is the entry paid tier, and the registry comment says so.

Why the entry tier and not the top of the ladder: a tenant brings its **own** Checkin.no/Tito account, and the integration reads that account through the tenant's own per-org credentials. There is no per-tenant cost to recover and no scarce platform resource to ration.

`badges` keeps `readiness: 'internal'` and **no** `minPlan`: issuance still signs with one global key pair, so until per-tenant signing keys exist (RunKonf/platform#46) a tier would sell something that cannot work.

#828's guard ("no platform-default feature may carry a `minPlan`") is **narrowed, not deleted**: it now asserts ticketing has the entry-tier `minPlan` and that every *other* platform-default feature (workshops, badges) stays internal and tier-less — so it still fails the moment someone attaches a tier to badges.

The platform org is unaffected by the tier: it is also a tenant and its own document may carry any plan (this deployment's carries none), so its implicit grant is asserted across every plan including community.

## 2. An explicit `enabled: false` is now a HARD kill switch

Before: a deny hid the nav entry, but the ticket pages resolved the provider first, so a deep link still rendered live sales for an org whose own credentials resolved. That is now closed at the page gate **and** in the dashboard action.

**Three states, ordered, and the order is load-bearing:**

1. explicit operator deny → **blocked** (nav, pages, dashboard tile; credentials do not rescue it, no provider call is made)
2. explicit grant / plan / platform org / own credentials → **allowed**
3. no decision + own credentials → **allowed** — the #828 guarantee, untouched

This refines #828's principle rather than reversing it, and both module docs now say so: the *absence* of a grant must never hide a surface that works (what provider-first protects); an *explicit deny* is a deliberate decision and is honoured everywhere. `isTicketingDeniedForOrg` is deliberately narrow — a nullish org, a missing document or a rejected Sanity read is **not** a deny, or one flaky read could blank a working page.

### Copy for a denied-but-credentialed org

A new `disabled` notice state, separate from `unavailable`: \"not available for your organization\" is true for a tenant that never had ticketing and a lie to one whose sales pages worked yesterday. The new copy is **\"Ticketing has been turned off for your organization\"** — it names what happened, says nothing was deleted and the tenant's own provider account is untouched, and points at the administrator. No settings link (there is nothing to fix). It deliberately does **not** name a vendor: unlike `unsupported`, this state does not depend on which provider the conference picked. The dashboard tile gets the matching `Turned Off` variant instead of telling a denied org to go and connect a provider. The tile now mirrors the resolver state for state (also from review): an org that is **not entitled at all** gets a `Not Available` variant rather than being nudged at settings it cannot use — a dead end that mattered little while ticketing was override-only and matters now that a plan sells it.

### Surfaces checked

- `/admin/tickets`, `/orders`, `/types`, `/discount`, `/companies` — all five go through `resolveTicketingAdminAccess`; all five blank on a deny.
- Admin nav / ⌘K — already honoured the deny via the registry decision; now covered by a test.
- Ticket-sales dashboard widget — its server action resolved the provider directly and would have kept streaming sales for a denied org; it now goes through the same three-state resolution.
- Badge surfaces (`/admin/speakers/badge`, the Manage Badges entry) already check the deny first and 404 — no change needed; the badges module doc now states the shared rule.
- `/admin/budget` — reads live ticket **income** straight from the provider, so a deny left it printing real revenue after every ticket page had gone dark (caught in review). Now resolved through the same gate; anything short of `ready` falls back to the manually-entered actuals, as an unconfigured provider always did.

Deliberately **not** gated: the public ticket-sales section (`src/lib/tickets/public.ts`), workshop eligibility and speaker-ticket issuance, and the status-page probe. Those are attendee-facing or backend flows, not organizer display surfaces — switching an organizer's admin view off should not break a live conference's public ticket sales. Flagging in case the owner wants a wider blast radius.

## Tests

`pnpm test`: **501 files / 7177 tests passing** (baseline on `main`: 500 / 7142 — +35 tests). `pnpm typecheck` clean, `npx prettier --check .` clean, `npx eslint .` 0 errors / 209 warnings (identical to the baseline).

Sabotage-proven, each guard removed by exact string, test failures observed, restored green:

| guard removed | tests that failed |
| --- | --- |
| the deny check in `resolveTicketingAdminAccess` | 11 (5 pages + provider-never-called + 4 resolver + dashboard action) |
| deny widened to \"not enabled\" (the #828 hazard) | 13 (incl. READY-on-rejected-read, and every unavailable state) |
| `minPlan: 'pro'` on ticketing | 18 |
| `readiness: 'ga'` reverted to internal | 5 |
| a `minPlan` added to **badges** | 1 (the narrowed guard still bites) |
| the platform-org grant in `ticketing.ts` | 15 |
| unresolvable org treated as a deny | 2 |
| the budget page's deny check | 1 |
| the tile's `unavailable` mapping | 1 |

Storybook: new `TicketingStateNotice` `Disabled` story and a `disabled (operator deny)` cell in the TicketSales matrix, both captured with `pnpm shoot` and inspected — no clipping, consistent with the neighbouring states.

<h3>Greptile Summary</h3>

This PR moves ticketing to the Pro plan and changes explicit ticketing denials into a deny-first kill switch.
- Adds a distinct disabled state and organizer-facing copy for ticket pages and the sales dashboard.
- Routes dashboard sales through the shared ticketing access resolver.
- Updates feature registry behavior and tests for paid-plan grants, platform defaults, and explicit denials.

The remaining budget-page bypass must be fixed before merging because a denied organization can still retrieve live ticket-income data there.

The new deny-first resolver works for migrated ticket pages and dashboard sales, but the budget page continues to resolve the provider directly and therefore does not honor the newly promised hard kill switch.

**Files Needing Attention:** src/lib/tickets/admin-access.ts and src/app/(admin)/admin/budget/page.tsx

| Filename | Overview |
|----------|----------|
| src/lib/tickets/admin-access.ts | Adds the deny-first disabled state correctly for its callers, but the centralized resolver is not used by the budget ticket-income surface. |
| src/app/(admin)/admin/actions.ts | Migrates dashboard ticket sales to the shared access resolver and returns a distinct disabled result before fetching provider data. |
| src/lib/features/platform-default.ts | Factors organization reads and adds a narrow explicit-deny query while preserving fail-closed ordinary entitlement behavior. |
| src/lib/features/registry.ts | Makes ticketing generally available at the Pro plan while preserving the platform organization’s separate implicit grant. |
| src/components/admin/TicketingStateNotice.tsx | Adds exhaustive disabled-state copy distinct from unavailable and unconfigured states. |
| src/components/admin/dashboard/widgets/TicketSalesDashboardWidget.tsx | Correctly renders the new disabled result as turned off rather than directing organizers to provider settings. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
  Request[Admin ticket surface] --> Deny{Explicit deny?}
  Deny -->|Yes| Disabled[Return disabled without provider call]
  Deny -->|No| Provider{Provider configured?}
  Provider -->|Yes| Ready[Fetch ticket data]
  Provider -->|No| Entitled{Entitled by plan, grant, or platform default?}
  Entitled -->|Yes| Unconfigured[Show connection guidance]
  Entitled -->|No| Unavailable[Show unavailable notice]
  Budget[Budget page] --> DirectProvider[Direct provider resolution]
  DirectProvider -. bypasses deny .-> Ready
```

<sub>Reviews (1): Last reviewed commit: ["feat(admin): sell ticketing at the entry..."](https://github.com/cloudnativebergen/website/commit/b9d8f15c91891a9447f758e9a4dc5a3a89a61b65) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=50496049)</sub>

**Context used:**

- Knowledge Base — [Admin Dashboard](https://app.greptile.com/cloudnativedays/-/custom-context/knowledge-base/cloudnativebergen/website/-/docs/admin-dashboard.md)

### DIFF
--- a/__tests__/app/admin/budget-ticketing-gate.test.tsx
+++ b/__tests__/app/admin/budget-ticketing-gate.test.tsx
@@ -1,0 +1,167 @@
+/**
+ * @vitest-environment node
+ *
+ * THE BUDGET PAGE IS A TICKETING SURFACE TOO. It shows live ticket INCOME,
+ * pulled straight from the conference's ticketing provider, so an operator's
+ * explicit ticketing deny has to reach it: a kill switch that blanks
+ * `/admin/tickets` while the budget page keeps printing live revenue is only
+ * half a switch.
+ *
+ * Everything the app owns runs for real — the provider resolver, the credential
+ * seam and the feature gate. Only Sanity, the provider HTTP client and the
+ * budget page's own client component are supplied.
+ */
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest'
+import { renderToStaticMarkup } from 'react-dom/server'
+
+const mockGetConference = vi.fn()
+const mockGetOrganizationById = vi.fn()
+
+vi.mock('@/lib/conference/sanity', () => ({
+  getConferenceForCurrentDomain: (...args: unknown[]) =>
+    mockGetConference(...args),
+}))
+
+vi.mock('@/lib/organization/sanity', () => ({
+  getOrganizationById: (...args: unknown[]) => mockGetOrganizationById(...args),
+  getOrganizationRefForCurrentConference: () => null,
+}))
+
+const h = vi.hoisted(() => ({
+  fetch: vi.fn(async () => null),
+  fetchEventTickets: vi.fn(),
+}))
+
+vi.mock('@/lib/sanity/client', () => ({
+  clientRead: { fetch: h.fetch },
+  clientReadUncached: { fetch: h.fetch },
+  clientWrite: { fetch: h.fetch },
+}))
+
+vi.mock('@/lib/tickets/provider/checkin', () => ({
+  CheckinProvider: class {
+    readonly name = 'Checkin.no'
+    isConfigured() {
+      return true
+    }
+    fetchEventTickets = h.fetchEventTickets
+  },
+}))
+
+vi.mock('@/lib/budget', () => ({
+  getBudgetForConference: vi.fn(async () => null),
+  deriveTicketIncome: vi.fn(() => ({ gross: 1000, net: 800, count: 4 })),
+  deriveManualTicketIncome: vi.fn(() => null),
+  deriveSponsorIncome: vi.fn(() => ({ total: 0, deals: [] })),
+}))
+
+vi.mock('@/lib/sponsor-crm/sanity', () => ({
+  listSponsorsForConference: vi.fn(async () => []),
+}))
+
+/** The client component is a marker: the gate decides before it renders. */
+vi.mock('@/components/admin', () => ({
+  BudgetPageClient: ({ ticketIncome }: { ticketIncome: unknown | null }) => (
+    <div data-live-ticket-income={ticketIncome ? 'yes' : 'no'} />
+  ),
+  ErrorDisplay: ({ title }: { title: string }) => <div>{title}</div>,
+}))
+
+import AdminBudgetPage from '@/app/(admin)/admin/budget/page'
+
+const PLATFORM_ORG_ID = 'org-platform'
+const TENANT_ORG_ID = 'org-A'
+
+function stubConference(orgId: string) {
+  mockGetConference.mockResolvedValue({
+    conference: {
+      _id: 'conf-1',
+      title: 'Tenant Conf',
+      checkinCustomerId: 42,
+      checkinEventId: 4242,
+      organization: { _ref: orgId, _type: 'reference' },
+    },
+    error: null,
+  })
+}
+
+/** Did the page hand live provider income to its client component? */
+async function renderedLiveIncome(): Promise<string | null> {
+  const markup = renderToStaticMarkup(
+    (await AdminBudgetPage()) as React.ReactElement,
+  )
+  return markup.match(/data-live-ticket-income="([^"]*)"/)?.[1] ?? null
+}
+
+beforeEach(() => {
+  vi.clearAllMocks()
+  vi.stubEnv('PLATFORM_ORG_ID', PLATFORM_ORG_ID)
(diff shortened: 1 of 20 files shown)


Starefossen:
Both threads predate the fixes and are addressed in the branch as it stands.

**Budget bypasses the kill switch** — fixed. The page now resolves through `resolveTicketingAdminAccess` rather than `resolveTicketingProvider` directly, so an explicit deny reaches it; everything short of `ready` falls back to manually-entered actuals. The comment records why: *a kill switch that leaves live ticket revenue on the budget page is only half a switch.* Adversarial QA verified it with `expect(h.fetchEventTickets).not.toHaveBeenCalled()`.

**The tile collapsing every non-ready state into `unconfigured`** — fixed. It now mirrors the resolver's state (`Not Available` / `Turned Off` / `Not Configured`), so a non-entitled community org is no longer told to "connect a provider in settings", which became a dead end once ticketing is sold by plan.

Separately: QA found two organizer-visible outputs the deny does NOT reach — the tickets tRPC router and the weekly Slack summary. Those are deliberately out of scope here and filed as **#836**, which asks whether the kill switch should mean the UI surface or every organizer-visible output. The module comment was corrected so it no longer claims "no deep link outlives it" — an API call is a deep link.

Issue #823, opened 2026-08-05
importAllHistoricSponsors is not organization-scoped


Starefossen:
`importAllHistoricSponsors` (`src/lib/sponsor-crm/sanity.ts:648`) queries `sponsorForConference` rows **with no organization filter**, so the historic-sponsor import offers one tenant's sponsors inside another tenant's import flow.

This is a tenancy bug independent of any demo work — it was found while auditing what a demo organization would leak into real tenants, but it applies to any two real tenants today. With one production organization it is currently inert; it becomes live the moment a second organization has sponsors.

Fix: scope the query to the requesting organization, the same way the rest of the sponsor CRM does. Worth checking the neighbouring queries in that module for the same shape while in there.

Related: the `no-unscoped-groq` lint rule did not catch this — worth checking whether the query shape is one of the known-invisible ones tracked in #792.

Issue #792, opened 2026-08-04
no-unscoped-groq: remaining invisible query shapes (predicate reorder, nested roots, _id in)


Starefossen:
`tenancy/no-unscoped-groq` guards a **shared multi-tenant Sanity dataset**, so a query shape the rule cannot see is a query shape nobody is checking. #782 closes two such shapes (whitespace `* [`, and `_id ==` root filters). This issue tracks what remains invisible.

Each shape below was confirmed by probe, not inferred from reading the patterns.

## Shapes the rule still cannot see

- `*[references($x)]` — no `_type`/`_id` token at all
- `*[slug.current == $slug]` — same
- **`_type`/`_id` not the first token**: `*[defined(foo) && _type == "x"]` — a plain predicate reorder walks around the rule
- Reversed comparison: `*["x" == _type]`
- `_type match`, and **`_id in $ids`** — note this narrows the #782 claim: only `_id ==` is closed, multi-id reads stay invisible
- **Nested roots inside a projection** — `GROQ_ROOT_FILTER.exec()` runs once and matches the outermost `*[`, so a nested `*[_id == ...]` is never examined
- Query built by split string concatenation

The predicate-reorder and nested-root cases are the ones I would rank highest: neither requires unusual style, and both are things a normal refactor produces by accident.

## Live census

A census of `src/` (non-test) found **exactly 9 rule-invisible root filters**, and **none is dangerous today**:

| Site | Shape | Why it is currently fine |
|---|---|---|
| `speaker/merge.ts:664` | `references($loserId)` | post-`requireExclusive` |
| `speaker/merge.ts:681` | `_id in $ids` | post-`requireExclusive` |
| `proposal/data/sanity.ts:406` | `references($proposalId)` | projects `_id,_type` only — delete-blocker enumeration |
| `gallery/sanity.ts:572` | `count(references($assetId))` | count only |
| `server/tenancy.ts` ×5 (196, 324, 419, 423, 463) | guard machinery | annotated, or carry in-query org predicates |

"None dangerous today" is a statement about the current call graph, not about the rule. The whitespace hole was also harmless right up until it wasn't — the point of closing these is that safety should not depend on nobody writing an ordinary predicate in an ordinary order.

## Suggested direction

[...]

PR #835, opened 2026-08-05
feat(tenancy): rewrite no-unscoped-groq as a parser-based, per-root-filter rule (P1)

Starefossen:
Implements **P1** of RunKonf/platform#48 — the engine, landed at `warn` with message parity. **P2 (ratchet + the newly-visible audit), P3 (kontroll port) and P4 (the tail) are NOT in this PR.** No `src/` file is touched.

## What was wrong

The rule anchored on the first `*[_type ==` / `*[_id ==` token per literal. Two defects, and the famous one is the smaller one:

- **Visibility** — predicate reorder, reversed comparison, `_type match`/`in`, `_id in $ids`, a filter opening on `references(...)` or `slug.current`, and structurally **every nested root filter**, because `exec()` stopped at the outermost `*[`.
- **Precision** — the rule *never inspected a root filter for a tenant predicate at all*. A correctly scoped hand-written query and a genuinely unscoped one produced byte-identical diagnostics. That is why #823 shipped after the rule had flagged both offending queries: true positives were indistinguishable from ~a quarter that were false.

## The engine as built

`eslint-rules/groq-scope-engine.js` — ESLint-free, options-driven, ready for the kontroll port.

1. **Flatten + substitute.** Each `${…}` gets a placeholder that parses, chosen *per hole* by hill-climbing on the parser's error offset (a parameter, a bare attribute, `_id,`, a number, a string, a predicate tail `&& $x`, a projection `{…}`, a slice `[0...1]`). Reusable **fragments** — a projection field list, a `&& …` predicate tail, a `,"reviews": *[…]` attribute — are wrapped so they become whole queries; the wrapper's synthetic root is discarded. Parameterised slice bounds (`[0...$limit]`), which groq-js rejects and Sanity accepts, are rewritten to constants **padded to the same length** so no offset moves.
2. **Enumerate.** Every `Everything`, filtered or bare (`count(*)` — D4), nested or not, with the root each is correlated to. Chained filters (`*[a][b]`) are one root constrained by `a && b`; a filter on a *non*-root (`items[…]` in a projection) contributes nothing.
3. **Judge each root independently**, recursively — which is the exact form of "every disjunct carries a tenant conjunct", and additionally gets `_type == "x" && (conference._ref == $conferenceId || organization._ref == $orgId)` right:

   ```
   scoped(a || b) = scoped(a) AND scoped(b)
   scoped(a && b) = scoped(a) OR  scoped(b)
   scoped(leaf)   = leaf is a tenant predicate (T1–T6)
   ```

   The vocabulary (T1–T6, the deliberate non-recognitions, the `_id ==` and identity axes) is **rule options**. A `kontroll`-shaped option set is exercised in the tests today.
4. **Builder modelling.** `scopedQuery` splices at `indexOf('*[')` and parenthesises what was there (verified in `src/lib/sanity/scoped.ts`); exactly the root at that offset is credited. Every other root — nested or not — is judged on its own.
5. **Per-root reporting and suppression.** The comment block above a literal governs the **first** root only. An outer annotation no longer vouches for nested roots it never saw.
6. **Fail closed.** A literal that looks like GROQ and does not parse is reported `unparseable`. Never silently clean.

Not checked, and said plainly in the rule header and the docs: **parameter provenance**. `conference._ref == $conferenceId` bound from client input — #826's actual bug — reports clean here, as under any static shape check.

## The walk-order claim: it held

groq-js 2.0.0 nodes carry no source spans, so the k-th `Everything` in walk order is paired with the k-th `*` token in the text. **I did not take the claim on faith.** Each pairing is *verified*: the root's own text is re-extracted and re-parsed, and the result compared structurally with the walked node. Any mismatch (or a count mismatch) invalidates the whole mapping, and the rule falls back to reporting at the literal — a fail-safe on *location*, never on the verdict.

Measured over every GROQ-shaped literal in `src/`: **327 of 327 verified reliable, 0 rejections.** Walk order and textual order coincide, as the PRD said.

Two findings from building the check:

- The verification is **falsifiable**, and it fires on a real shape: a filter chained onto a parenthesised or piped root (`(*[_type == "a"])[conference._ref == $conferenceId]`) puts the constraining predicate *outside* the `*[…]` the token opens. Count still matches; only the re-parse catches it. Both a rule test and an engine test pin it.
- Building the star scanner surfaced a bug the verification caught for me: `_id in *[_type == "conference"]…` was read as "identifier, then multiplication" and the root was missed. The scanner now tracks the last significant *token*, not the last character.

## One false negative found in adversarial QA, fixed here

`${prefix}*[_type == "talk"]` reported **CLEAN** — the one shape where this rewrite was *quieter* than the regex it replaced, which is the exact failure mode that would make it a net loss.

After substitution the text is `$__groqInterp0*[_type == "talk"]`, which is valid GROQ: a parameter **multiplied by an array literal**. No `Everything` node is created, so there is no root to judge. The star scanner independently agrees (a `*` after an operand is multiplication), the counts match, and the location mapping verifies — every part of the analysis correctly agrees there is no root, on a query that reads every tenant at runtime. The old `/\*\s*\[\s*_(type|id)\s*==/` flagged it, because it did not care what preceded the `*`.

**Fix:** any `*` separated from an interpolation by whitespace alone, which the scanner did not count as an `Everything`, is reported `interpolatedFilter`. The test is **placeholder-ness**, not the placeholder's shape — every rung the ladder emits ends in an operand (`$p`, `_id`, `0`, `"…"`, `{…}`, `[0…1]`), so a future rung inherits the guard for free. The `[0...1]` rung is closed by the other fail-closed path (a leading slice is not an expression, so the literal is `unparseable`); both branches are pinned.

Reported **regardless of `scopedFetch`**: `scopedQuery` splices at the first `*[` of the *assembled* string, and the prefix is exactly the text neither it nor this rule can read.

`rg -U '\$\{[^}]*\}\s*\*\s*\['` over non-test `src/` returns **0 hits** and the census is unchanged at 201 — latent, and future code is the point of a lint rule. Six rule cases and three engine cases pin it, and the guard is sabotage-verified twice (removing the detection: 7 failures; letting the builder silence it: 1).

## A second false negative of the same class, also fixed

`insideBuilder` returned true for a literal nested **anywhere** inside a `scopedFetch(...)` call — it walked up looking only for an enclosing call. But `scopedFetch(client, scope, groqBody, params?, options?)` splices into `groqBody` and nothing else:

```ts
scopedFetch(client, { orgId }, mainQuery, {
  ids: idsFrom(`*[_type == "talk"]`),   // credited, wrongly — reported CLEAN
})
```

Confirmed by probe for **all four** non-query positions (client, scope, params, options), each reporting clean. Same class as the swallowed star: something looks covered because a *container* vouches for it.

**Fix:** crediting is tied to the query **argument**. Within it a literal still counts through a ternary or a `??` default — either could BE the value — but not through a function call (`wrap(…)`), whose return value is what gets spliced; that falls back to being judged on its own, exactly like a body passed by variable. The index is a rule option (`builderQueryArg`, default 2, verified against the real signature) beside `builderName`, so a different builder is configured rather than forked.

**No behaviour change on existing code:** 22 real `scopedFetch` call sites, census unchanged at 201, and the full diagnostic set (`file:line:column:messageId`) is byte-identical before and after — every site already passes its query inline in the right position. Two sabotage entries cover it: dropping the position check fails 2 tests, dropping the value-preserving-path check fails 1 (they are complementary — position catches direct arguments, the path check catches nested ones).

## Two smaller items, stated rather than silently skipped

- **`AC7` (the #823 before/after pair as a rule test) is DEFERRED to P2.** P1's gate is AC1–6 + 9; AC7 is listed here so P2 inherits it as a known item rather than a surprise. Both #823 queries are ordinary `_type ==`-first roots — the class this engine already covers — so the test is cheap; it just is not written.
- **A literal that is nothing but `*` is not flagged**, though D4 says a bare `*` reads every tenant. `count(*)`, `*{…}` and `* | order(…)` **are** flagged. Widening the hint gate is not the one-liner it looks like: `src/` carries ~14 bare-`*` literals and **every one is a CORS `Access-Control-Allow-Origin` or a `robots.txt` user-agent**. Flagging them would add pure noise to a control whose whole value is that its output is trustworthy — the problem this PR exists to fix. Telling them apart needs the literal's *context* (is it an argument to a query call?), which is a rule-level design question. The predecessor was equally blind, so nothing regressed. Documented in the rule and in `TENANT_SCOPING.md`; left for P2.

## Census, re-derived on this branch (method: `eslint -f json`, then a per-literal re-run of both rules through `Linter`)

| | before (`107eaf95`) | after |
|---|---|---|
| total | **209** | **201** |
| `unscoped` | 201 | 171 |
| `interpolatedFilter` | 8 | 28 |
| `unparseable` | — | 2 |

Baseline reproduces the PRD's 209/201/8 exactly. The delta accounted for, per literal:

- **37 literals went from flagged to entirely clean.** 37 of the 38 cleared diagnostics sit on literals carrying a canonical tenant predicate — the false positives the PRD counted as 47 under definition (a).
- **10 literals went from clean to flagged** (4 flat, 6 nested-root).
- **28 of the 201 diagnostics sit on nested/subsequent roots.** Every one was structurally invisible before — most on literals that were *already* noisy, so the literal was reported but the nested root inside it was never examined.
- The remaining false-positive gap is honest and explains itself: **12 of the ~49 canonical-predicate literals still report**, because the root filter also carries an interpolation (`status != "${Status.draft}"`, `${statusFilter}`). §4 lists that among the deliberate non-recognitions — injected text can escape the bracket — so they moved from `unscoped` to `interpolatedFilter`, which is why that count went 8 → 28. Same verdict, better message; P2 decides annotate-or-parameterise.
- **`interpolatedFilter` semantics widened deliberately:** it now fires for an interpolation *anywhere* in a root filter, not only a leading one. Before, `*[_type == "x" && ${extra}]` was reported `unscoped`; a rule that recognises tenant predicates must not let one vouch for text it cannot see.
- Only **2 literals** are `unparseable` (`src/lib/proposal/data/sanity.ts:76, :210` — nested ternaries assembling four fragments). The ladder + wrappers resolve the other 50 that a naive substitution left broken.

## The P2 work list (do not fix these here)

**Newly flagged literals — clean before, flagged now (10):**

```
src/lib/gallery/sanity.ts:572          count(*[references($assetId)])
src/lib/proposal/data/sanity.ts:406    *[references($proposalId) && _id != $proposalId]
src/lib/speaker/merge.ts:685           *[references($loserId) && _id != $loserId]
src/lib/speaker/merge.ts:702           *[_id in $ids]
src/lib/sponsor-crm/bulk.ts:295        nested: contractDocument.asset._ref == ^._id
src/lib/sponsor/sanity.ts:291          nested: sponsor._ref == $id → .conference->organization._ref
src/lib/sponsor/sanity.ts:343          nested: contractDocument.asset._ref == ^._id
src/lib/travel-support/sanity.ts:122   nested: travelExpense … travelSupport._ref == ^._id
src/server/tenancy.ts:359              nested ×2: foreignTalkCount / foreignRefCount
src/server/tenancy.ts:558              count(*[_id in $ids && _type == $expectedType && …])
```

Six of these were named "invisible but not dangerous" in the #792 census (`merge.ts`, `gallery:572`, `proposal:406`, `tenancy.ts`) — they now say so out loud instead of relying on a doc.

**All 28 nested-root diagnostics, grouped by file**, are below. Classified:

- **19** are parent correlations of the form `<field>._ref == ^._id` (`"expenses": *[… travelSupport._ref == ^._id]`, `"signupCount": count(*[… workshop._ref == ^._id])`). **This is the single biggest P2 decision, and it is not mine to make.** T6 as the PRD specifies it recognises `conference._ref == ^.conference._ref`; the general principle — a correlation to a root that is *itself* proven scoped — would cover `^._id` too, and the engine already tracks `parentScoped`. Recognising it would clear most of this set mechanically. I did **not** widen the vocabulary unilaterally: it is a security-relevant judgement, and PRD §9 has no decision for it. Suggest filing it as **D8** before the P2 burn-down.
- **6** are `references(^._id)` correlations (same question).
- **3** are genuine platform aggregates or `_id` reads.

<details><summary>The 28, grouped by file</summary>

**src/app/(admin)/admin/actions.ts**
- L1023 (root #1, `unscoped`) — `"reviews": *[_type == "review" && proposal._ref == ^._id]{ score }`

**src/lib/featured/sanity.ts**
- L181 (root #1, `unscoped`) — `"featuredTalksCount": count(*[_type == "conference" && _id == $conferenceId][0].featuredTalks),`
- L182 (root #2, `unscoped`) — `"availableSpeakersCount": count(*[_type == "speaker" && count(*[_type == "talk" && references(^._id) && c`

**src/lib/messaging/sanity.ts**
- L763 (root #1, `unscoped`) — `"lastMessage": *[_type == "message" && conversation._ref == ^._id] | order(createdAt desc) [0] {`

**src/lib/proposal/data/sanity.ts**
- L93 (root #1, `unscoped`) — `"reviewCount": count(*[_type == "review" && proposal._ref == ^._id]),`
- L94 (root #2, `unscoped`) — `"reviewScores": *[_type == "review" && proposal._ref == ^._id]{"total": score.content + score.relevance +`
- L109 (root #1, `unscoped`) — `"reviewCount": count(*[_type == "review" && proposal._ref == ^._id]),`
- L110 (root #2, `unscoped`) — `"reviewScores": *[_type == "review" && proposal._ref == ^._id]{"total": score.content + score.relevance +`
- L256 (root #1, `unscoped`) — `"waitlistCount": count(*[_type == "workshopSignup" && workshop._ref == ^._id && status == "waitlist"]),`
- L257 (root #2, `unscoped`) — `"available": coalesce(capacity, 30) - count(*[_type == "workshopSignup" && workshop._ref == ^._id && stat`
- L817 (root #1, `unscoped`) — `"coSpeakerInvitations": *[_type == "coSpeakerInvitation" && proposal._ref == ^._id]{`

**src/lib/speaker/sanity.ts**
- L870 (root #3, `interpolatedFilter`) — `"schedule": *[_type == "schedule" && conference._ref == $conferenceId && ${OFFICIAL_SCHEDULE_FILTER} && ^`
- L1039 (root #1, `interpolatedFilter`) — `const query = groq'*[_type == "speaker" && count(*[_type == "talk" && references(^._id) && status in [${s`
- L1044 (root #2, `interpolatedFilter`) — `"proposals": *[_type == "talk" && references(^._id) && status in [${statusFilter}] ${proposalsConferenceF`

**src/lib/sponsor-crm/bulk.ts**
- L298 (root #1, `unscoped`) — `count(*[_type == "sponsorForConference" && contractDocument.asset._ref == ^._id && !(_id in $ids)]) == 0`

**src/lib/sponsor-crm/sanity.ts**
- L136 (root #1, `unscoped`) — `"activityCount": count(*[_type == "sponsorActivity" && sponsorForConference._ref == ^._id])`
- L339 (root #1, `unscoped`) — `count(*[_type == "sponsorForConference" && contractDocument.asset._ref == ^._id && _id != $id]) == 0`

**src/lib/sponsor/sanity.ts**
- L292 (root #1, `unscoped`) — `"linkedOrgs": *[_type == "sponsorForConference" && sponsor._ref == $id].conference->organization._ref`
- L346 (root #1, `unscoped`) — `count(*[_type == "sponsorForConference" && contractDocument.asset._ref == ^._id && !(sponsor._ref == $id)`

**src/lib/system-status/checks.ts**
- L860 (root #1, `unscoped`) — `"speakers": count(*[_type == "speaker" && count(pushSubscriptions) > 0])`

**src/lib/travel-support/sanity.ts**
- L34 (root #1, `unscoped`) — `"expenses": *[_type == "travelExpense" && travelSupport._ref == ^._id] | order(_createdAt desc) {`
- L76 (root #1, `unscoped`) — `"expenses": *[_type == "travelExpense" && travelSupport._ref == ^._id] | order(_createdAt desc) {`
- L133 (root #1, `unscoped`) — `"expenses": *[_type == "travelExpense" && travelSupport._ref == ^._id] {`

**src/lib/workshop/sanity.ts**
- L69 (root #1, `unscoped`) — `"signupCount": count(*[_type == "workshopSignup" && workshop._ref == ^._id && status == "confirmed"])`
- L233 (root #1, `unscoped`) — `"signupCount": count(*[_type == "workshopSignup" && workshop._ref == ^._id && status == "confirmed"])`
- L358 (root #1, `unscoped`) — `"signups": *[_type == "workshopSignup" && workshop._ref == ^._id] {`

**src/server/tenancy.ts**
- L362 (root #1, `unscoped`) — `"foreignTalkCount": count(*[_type == "talk" && references(^._id) && defined(conference->organization._ref`
- L363 (root #2, `unscoped`) — `"foreignRefCount": count(*[references(^._id) && _id != ^._id && (!defined(coalesce(organization._ref, con`
</details>

## Tests

- `eslint-rules/no-unscoped-groq.test.ts` — **flipped into characterization tests for the new behaviour, not deleted.** The two cases that used to assert *silence* on a nested root inside `scopedFetch`, and on an outer annotation covering a nested root, now assert the diagnostic, with the comment explaining that the fix flipped a documented expectation. Whitespace (`* [`) and `_id ==` regression cover kept verbatim — they now pin that the parser inherits the property for free.
- Acceptance criteria are encoded as tests: **AC1** the four-probe file yields exactly one diagnostic; **AC2** every #792 invisible shape, written unscoped, is flagged; **AC3** one scoped + one unscoped root ⇒ one diagnostic on the unscoped root's line, and an outer annotation does not reach a nested root; **AC4** inside `scopedFetch` the first root is clean and an unscoped nested root is still flagged; **AC5** fail-open flagged and `groq-global-scoped:` does not clear it; **AC6** an unparseable literal is reported.
- `eslint-rules/groq-scope-engine.test.ts` — new: the location mapping (15 shapes incl. nested, sibling, function-argument, `in`-keyword, multiplication and string-literal decoys), node-for-node order agreement, slice normalisation, T1–T6, the non-recognitions, the disjunct rule, and kontroll's vocabulary.

**Sabotage: 20 behaviours mutated, 20 detected.**

| mutation | failing tests |
|---|---|
| T1/T3 tenant-ref recognition | 25 |
| chained filter contributes its predicate | 38 |
| nested roots are enumerated at all | 24 |
| per-root reporting (first root only) | 9 |
| builder credits only the root it splices into | 11 |
| fail-open `!defined(tenant)` detection | 7 |
| a `*` swallowed by a substitution is reported | 7 |
| interpolation taints its root | 7 |
| recursive disjunct rule (`Or` behaves like `And`) | 6 |
| `in`/`match` keyword handling in the star scanner | 3 |
| location-mapping verification (re-parse each root) | 3 |
| per-root suppression (literal annotation reaches nested roots) | 2 |
| masked-star reporting is not silenced by the builder | 1 |
| the builder credits only its QUERY argument | 2 |
| only value-preserving parents keep the credit | 1 |
| fail CLOSED on an unparseable literal | 2 |
| the GROQ-query gate (parse every string) | 2 |
| T6 requires a SCOPED parent | 1 |
| a filter on a NON-root must not credit the root | 1 |
| slice-bound normalisation | 1 |

## Numbers

- `pnpm test` — **501 files, 7243 tests, all passing** (baseline `107eaf95`: 500 files / 7142 tests; `eslint-rules/` went 1 file / 63 tests → 2 files / 164 tests).
- `pnpm typecheck` — clean.
- `npx prettier --check .` — clean.
- `npx eslint .` — **209 → 201** `tenancy/no-unscoped-groq` warnings, 0 errors, no other rule's output changed.
- **AC9, lint wall time:** baseline 17.14 / 15.35 / 15.53 s; this branch 15.98 / 16.42 / 17.19 s. Noise-level, for ~344 literals parsed (367 root filters enumerated).

## Docs

`docs/TENANT_SCOPING.md`'s lint-rule section is rewritten to describe the rule that exists: the five shapes, the T1–T6 table, the deliberate non-recognitions, the provenance limitation, how a template literal is read, and per-root suppression. The stale censuses ("9 rule-invisible sites", "26 literals / 37 nested roots") are **deleted** rather than marked superseded — the doc states current behaviour, and the numbers live in this PR. The full AC10 rewrite belongs to P2, after the audit.


Starefossen:
All three review points are addressed; resolving.

**`insideBuilder` over-credited nested literals** — reproduced, and **worse than reported**: probing all five `scopedFetch` argument positions showed *every* argument was credited, not just `params`. Crediting is now tied to the query **argument** (index read from `scoped.ts`, not assumed) with two complementary guards: the path must start at that argument, and within it the literal must be able to *be* the value — surviving a ternary, `??` or a TS assertion, but **not** a function call, since `scopedQuery` splices into whatever the function returns. The nearest builder call decides; an outer one cannot rescue it. The argument index is a rule option so kontroll's signature configures rather than forks.

Verified as latent, not shipped: 22 real call sites, and the full diagnostic set (`file:line:column:messageId`) is **byte-identical** before and after — not merely the same total.

**`meta.messages.unscoped`** — the metadata string itself now enumerates what the rule actually reports, including nested roots and bare `*`.

**`rg -c`** — done; no `grep -c` remains in the doc.

Sabotage is now **20 mutations, 20 detected**. Two pre-existing mutations needed re-pointing after the rename, and one of those was silently passing as PATTERN-NOT-FOUND — a sabotage that "passed" because its pattern never matched. That one is honest now too.

PR #840, opened 2026-08-05
fix(admin): stop four fresh-tenant admin surfaces from stating things that are not true

### LINKED ISSUE #838: Fresh-tenant admin shows wrong values, not empty ones: NaNd, a CFP reported "Open", a silent no-op, a fabricated date
A freshly provisioned conference has no `cfpStartDate`/`cfpEndDate`, and three admin surfaces respond by displaying wrong information rather than no information. Found in an onboarding review against SaaS empty-state practice, where the relevant rule is that a broken screen is worse than a blank one.

## 1. CFP Health widget renders `NaNd`

`src/components/admin/dashboard/widgets/CFPHealthWidget.tsx` (~82-100) computes days-until from an absent `cfpStartDate` with no guard, so the tiles read **"Opens In: NaNd"** and **"Duration: NaNd"**.

## 2. Proposal Pipeline reports the CFP is "Open"

`src/components/admin/dashboard/widgets/ProposalPipelineWidget.tsx` (~45-95) derives CFP status from the same absent dates. `NaN > 0` is false, so the widget concludes the CFP is **Open** on a conference whose CFP was never configured — an affirmative false statement, not a rendering artefact.

Neither is caught by the widget matrix stories, because those fixtures always supply dates.

## 3. Schedule editor silently discards input

`/admin/schedule` → "Create First Track" accepts a track name and does nothing. With no conference dates `generateConferenceDates()` returns `[]`, and `src/lib/schedule/reducer.ts:144` returns state unchanged (`if (!current) return state`). No error, no explanation. It is reachable directly from the dashboard's "Configure Schedule" CTA, so it is on the likely first-session path.

Should say "Set your conference dates first" with a link to settings.

## 4. Marketing page fabricates a date

The announcement card falls back to a hardcoded `'June 15, 2025'` when `startDate` is absent — on a **downloadable share asset**, so the invented date can leave the product and reach an audience.

## Why these are grouped

All four are the same shape: an unconfigured field rendered as a confident value. #830 fixed this class on the public site (blank `<h1>`, "Meet our 0 speakers", `/info` asserting an afterparty); this is the admin half, and the marketing asset can escape the admin entirely.

Starefossen:
Closes #838.

A conference provisioned by `buildOnboardingDocuments` has **no dates at all** — no `startDate`/`endDate`, no `cfpStartDate`/`cfpEndDate`. Four admin surfaces answered that silence with a confident wrong value instead of nothing. #830 fixed this class on the public site; this is the admin half.

## Each surface, before → after

| Surface | Before | After |
| --- | --- | --- |
| `CFPHealthWidget` (initialization view) | Tiles read **"Opens In: NaNd"**, **"Duration: NaNd"** — `new Date(undefined).getTime()` is `NaN` | House `WidgetEmptyState`: "No CFP dates set yet" + a **Set CFP dates** link to `/admin/settings` |
| `ProposalPipelineWidget` (initialization/planning view) | **"CFP Opens: Open"** — `NaN > 0` is false, so the widget affirmatively stated a never-configured CFP was accepting submissions; Duration read `NaNd` | Same empty state, same link |
| `/admin/schedule` "Create First Track" | Modal accepted a track name and **silently discarded it** — zero conference dates means zero days, and the reducer's `if (!current) return state` dropped it. Reachable from the dashboard's "Configure Schedule" CTA | The editor renders **"Set your conference dates first"** with a link to settings and offers no track-creation affordance at all (desktop *and* mobile) |
| `/admin/marketing` conference promo card | Hardcoded **`'June 15, 2025'`** on a *downloadable* share graphic | Date line **omitted entirely**; `SponsorThankYou` already treats `eventDate` as optional |

## Judgement calls

**1. The marketing card omits the date line rather than showing a placeholder.**
Agreed with the instinct in the issue. This is a share asset that leaves the product: an invented date is a lie, and "Date TBA" printed on a social graphic is a defect an organizer would have to notice and work around. Omitting collapses the metadata row to just the location chip, which reads as a deliberate design rather than a hole. (The adjacent `'Location TBA'` fallback is left alone — it is honest, not fabricated, and changing it is not this issue.)

**2. The silent no-op is fixed at BOTH tiers, with the UI tier doing the load-bearing work.**
The issue asked for whichever makes the discard impossible rather than merely explained, so:
- **Impossible (UI):** `ScheduleEditor` bails to `NoConferenceDatesState` when `state.schedules.length === 0`, before either the desktop board or the mobile view mounts. The `AddTrackModal` is unreachable, so there is no input to discard. `schedules.length === 0` is exactly equivalent to "no usable conference dates" — `getScheduleData()` emits one day per date and always emits at least one when the dates are valid.
- **Explained (reducer):** the ten `if (!current) return state` no-ops now return `withoutActiveDay(state)`, which sets `ui.error` (surfaced by the existing `ErrorBanner`). This is the backstop: any *future* path that dispatches an edit with no day loaded gets a visible error instead of a silent drop.

**3. Decorative empty-state icons are hidden at the SLOT, not per call site** (from review). All three shared empty-state components rely on Heroicons' default `aria-hidden="true"` — a convention, not a guarantee, since `WidgetEmptyState.icon` is an arbitrary `ReactNode`. The attribute now sits on the icon slot in `WidgetEmptyState`, covering every current and future caller. `EmptyState.tsx` / `DataTable/TableEmptyState.tsx` take a `ComponentType` and are untouched by this PR, so they were left alone.

## Why the tests missed it, and what now catches it

Every `conferenceInPhase()` matrix fixture supplies a full set of dates, so no cell ever rendered the day-one shape. Added:

- `freshlyProvisionedConference()` in `__matrix__/fixtures.ts`, built from the **real** `buildOnboardingDocuments` (same approach as #824/#830), plus day-one cells in the CFPHealth and ProposalPipeline matrices and a `NoConferenceDates` story on `ScheduleEditor` — whose fixture also calls the real builder, so no hand-written approximation exists to drift.
- `__tests__/app/day-one/fresh-tenant-admin-surfaces.test.tsx` — the admin counterpart to `fresh-tenant-public-pages.test.tsx`. Its **premise guard** asserts what is genuinely absent *today*: `startDate`, `endDate`, `cfpStartDate`, `cfpEndDate`, `cfpNotifyDate`, `programDate` — and asserts that `formats` **is** seeded, since #833 changed that. If provisioning drifts, the premise fails loudly instead of the coverage quietly evaporating.
- A `WidgetEmptyState` a11y contract test in `widget-shared.test.tsx` that passes a bare `<svg>`, so the slot-level guard is exercised rather than masked by Heroicons' own default.

Follow-up noted in-code: this is now the third story-side copy of the provisioning call (here, `InfoContent.stories.tsx`, `__matrix__/fixtures.ts`). #841 extracts the test-side equivalent to `__tests__/testdata/onboarding.ts`; the story-side copies should collapse onto one helper once that lands (it is not on `main` yet).

## Sabotage results

Each guard removed by exact string, suite re-run, restored, re-run green. **7 guards, 10 failing-test outcomes.**

| Guard removed | Failing tests |
| --- | --- |
| `CFPHealthWidget` unconfigured branch | 2 |
| `ProposalPipelineWidget` unconfigured branch | 2 (failure output confirms the literal `Open` rendering) |
| `ScheduleEditor` `schedules.length === 0` early return | 2 |
| reducer `addTrack` → `withoutActiveDay(state)` | 1 |
| marketing `: undefined` → `: 'June 15, 2025'` | 2 |
| marketing `{eventDate && …}` wrapper | 1 |
| `WidgetEmptyState` icon-slot `aria-hidden` | 1 |

## Numbers (re-derived on this branch)

| Check | Baseline (`origin/main`) | This branch |
| --- | --- | --- |
| `pnpm test` | 503 files / 7243 tests, all pass | **504 files / 7258 tests**, all pass |
| `npx eslint .` | 0 errors, 209 warnings | 0 errors, 209 warnings |
| `pnpm knip` | clean | clean |
| `pnpm build-storybook` | — | completes successfully |

## Visual inspection — not done, and why

`pnpm shoot` could not run here: **every** Chromium build on this machine (`chromium_headless_shell-1234`, `chromium-1234`, and a manual launch of `Google Chrome for Testing`) exits with `SIGTRAP` moments after printing `DevTools listening`, whether launched by Playwright, by hand, or under a pty. This is an environment failure, not a change in this PR — a bare `chromium.launch()` fails identically on unmodified code. Storybook itself builds and serves fine.

Compensating cover: the new states use the existing house components (`WidgetEmptyState`, which already carries the widget height contract), `pnpm build-storybook` passes, Storybook Tests and Visual Regression Tests are green in CI, and the new day-one cells are in the matrices ready to be looked at. **Please eyeball the three new cells before merging**: `Systems/Proposals/Admin/Dashboard/Matrix/CFPHealth → Phases`, `…/ProposalPipeline → Phases`, and `Systems/Program/Admin/ScheduleEditor → NoConferenceDates`.

### DIFF
--- a/__tests__/app/day-one/fresh-tenant-admin-surfaces.test.tsx
+++ b/__tests__/app/day-one/fresh-tenant-admin-surfaces.test.tsx
@@ -1,0 +1,304 @@
+/**
+ * @vitest-environment jsdom
+ *
+ * DAY ONE, ADMIN SIDE — the counterpart to `fresh-tenant-public-pages.test.tsx`
+ * (#830), which covered the public half of the same class of bug.
+ *
+ * `@/lib/onboarding/create.ts` provisions a tenant with a title, an org
+ * reference, a city/country, contact addresses and a starter format list —
+ * and NO dates of any kind: no `startDate`/`endDate`, no `cfpStartDate`/
+ * `cfpEndDate`. Four admin surfaces responded to that silence by stating
+ * something false rather than nothing (#838):
+ *
+ *   - the CFP Health widget subtracted an absent `cfpStartDate` from now and
+ *     printed the result, so its tiles read "Opens In: NaNd" / "Duration: NaNd";
+ *   - the Proposal Pipeline widget derived CFP status from the same absent
+ *     dates, and because `NaN > 0` is false it reported the CFP as **Open** on a
+ *     conference whose CFP had never been configured;
+ *   - `/admin/schedule` offered "Create First Track", took a track name, and
+ *     dropped it — with zero conference dates there are zero days, so the
+ *     reducer had nothing to attach it to and returned state unchanged;
+ *   - `/admin/marketing` printed a hardcoded 'June 15, 2025' on a DOWNLOADABLE
+ *     share graphic, so the invented date could leave the product entirely.
+ *
+ * Every conference here is built by the REAL `buildOnboardingDocuments`, so if
+ * provisioning ever starts seeding dates the premise guard below fails loudly
+ * instead of this suite quietly testing a scenario that no longer exists.
+ */
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest'
+import { render, screen, cleanup, fireEvent } from '@testing-library/react'
+import type { ReactElement } from 'react'
+
+/* -------------------------------------------------------------------------- */
+/* Module boundaries                                                          */
+/* -------------------------------------------------------------------------- */
+
+const fetchCFPHealthMock = vi.fn()
+const fetchProposalPipelineMock = vi.fn()
+vi.mock('@/app/(admin)/admin/actions', () => ({
+  fetchCFPHealth: () => fetchCFPHealthMock(),
+  fetchProposalPipeline: () => fetchProposalPipelineMock(),
+}))
+
+const {
+  saveMutateAsync,
+  actionMutateAsync,
+  pollInvalidate,
+  pollSetData,
+  routerRefresh,
+} = vi.hoisted(() => ({
+  saveMutateAsync: vi.fn(),
+  actionMutateAsync: vi.fn(),
+  pollInvalidate: vi.fn(),
+  pollSetData: vi.fn(),
+  routerRefresh: vi.fn(),
+}))
+
+vi.mock('@/lib/trpc/client', () => ({
+  api: {
+    useUtils: () => ({
+      schedule: {
+        admin: {
+          pollVersions: { invalidate: pollInvalidate, setData: pollSetData },
+        },
+      },
+    }),
+    schedule: {
+      save: { useMutation: () => ({ mutateAsync: saveMutateAsync }) },
+      action: { useMutation: () => ({ mutateAsync: actionMutateAsync }) },
+      admin: { pollVersions: { useQuery: () => ({ data: undefined }) } },
+    },
+  },
+}))
+
+vi.mock('next/navigation', () => ({
+  useRouter: () => ({ refresh: routerRefresh }),
+}))
+
+/* Marketing page boundaries. */
+const getConferenceForCurrentDomainMock = vi.fn()
+vi.mock('@/lib/conference/sanity', () => ({
+  getConferenceForCurrentDomain: (...args: unknown[]) =>
+    getConferenceForCurrentDomainMock(...args),
+}))
+vi.mock('@/lib/auth', () => ({ getAuthSession: vi.fn(async () => ({})) }))
+vi.mock('@/lib/authz/organizer', () => ({
+  isOrganizerForCurrentOrg: vi.fn(async () => true),
+}))
+vi.mock('@/lib/gallery/sanity', () => ({
+  getFeaturedGalleryImages: vi.fn(async () => []),
+}))
+vi.mock('@/lib/proposal/server', () => ({
+  getProposals: vi.fn(async () => ({ proposals: [], proposalsError: null })),
+}))
+vi.mock('@/components/CloudNativePattern', () => ({
+  CloudNativePattern: () => <div data-testid="pattern" />,
+}))
+vi.mock('@/components/admin/MemeGeneratorWithDownload', () => ({
+  MemeGeneratorWithDownload: () => <div data-testid="meme-generator" />,
+}))
+vi.mock('@/components/admin/PhotoGalleryWithDownload', () => ({
+  PhotoGalleryWithDownload: () => <div data-testid="photo-gallery" />,
+}))
+vi.mock('@/components/common/DownloadableImage', () => ({
+  DownloadableImage: ({ children }: { children: React.ReactNode }) => (
+    <div data-testid="downloadable">{children}</div>
+  ),
+}))
+
+import { buildOnboardingDocuments } from '@/lib/onboarding/create'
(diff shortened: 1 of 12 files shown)


Starefossen:
Re the Greptile a11y notes on the three Calendar icons: Heroicons v2 React components already render `aria-hidden="true"` on the SVG root by default (verified in `node_modules/@heroicons/react/24/outline/esm/CalendarIcon.js`), and the surrounding house usages rely on that same default. No change made.

The fixture-comment note was valid and is fixed in 8d29dc4 — the ScheduleEditor story stub is now described as a hand-written stand-in rather than implying provisioning parity, with a pointer to the provisioning-sourced fixture that actually guards the premise.

Starefossen:
All four review threads addressed in 80b741e2 (plus 8d29dc4).

**Decorative icons (3 threads) — fixed centrally, not per call site.**
Checked the shared components first, as asked. Three exist: `WidgetEmptyState` (`widgets/shared.tsx`), `EmptyState.tsx` and `DataTable/TableEmptyState.tsx`. None sets `aria-hidden` explicitly — all three rely on Heroicons, which *do* emit `aria-hidden="true"` on the SVG root by default (verified in `node_modules/@heroicons/react/24/outline/esm/CalendarIcon.js`, not assumed). So the flagged icons were already hidden.

That is a convention, not a guarantee: `WidgetEmptyState` takes an arbitrary `ReactNode` icon, so a caller can pass a hand-rolled SVG that hides nothing. I put `aria-hidden="true"` on the icon **slot** in `WidgetEmptyState` — structural, covers both widget states in this PR and every future one for free. `EmptyState`/`TableEmptyState` take a `ComponentType` and render it themselves, and neither is touched by this PR, so I left them; if you want the same slot-level guard there it is a one-liner each, say the word.

The schedule editor's state is not a `WidgetEmptyState`; its two Heroicons already carry the attribute. Rather than sprinkle redundant attributes I asserted the **outcome** on all three new day-one states (no `<svg>` reaches the a11y tree), plus a `WidgetEmptyState` unit test that passes a bare `<svg data-testid>` so the structural guard is genuinely exercised. Sabotage: removing `aria-hidden` from the slot fails that unit test (1 failure); the outcome tests stay green, because Heroicons self-hide — which is exactly why the slot-level guard was worth adding.

**Fixture comment — the approximation is gone, not just reworded.**
`__tests__/testdata/onboarding.ts` from #841 is **not** on `origin/main` yet (main is at 3bb31713; `__tests__/testdata/` has conference/proposals/speakers only). But the cheap fix was available anyway: the ScheduleEditor story now calls the real `buildOnboardingDocuments`, exactly as `InfoContent.stories.tsx` does from #830. There is no hand-written stub left, so nothing can claim parity it does not have. The consolidation follow-up (this is now the third story-side copy — here, `InfoContent.stories.tsx`, `widgets/__matrix__/fixtures.ts`) is noted in the comment where the duplication actually lives.

**Gates re-run on 80b741e2**

| Check | Baseline (`origin/main`) | This branch |
| --- | --- | --- |
| `pnpm test` | 503 files / 7243 tests | **504 files / 7258 tests**, all pass |
| `npx eslint .` | 0 errors, 209 warnings | 0 errors, 209 warnings |
| `pnpm knip` | clean | clean |
| `pnpm build-storybook` | — | completes successfully |

Sabotage count is now 7 guards, 10 failing-test outcomes. Still not merging.

PR #847, opened 2026-08-05
feat(ticketing): make an operator deny a kill switch for every organizer-visible output

### LINKED ISSUE #836: Decide the ticketing kill switch's scope: UI surface, or every organizer-visible output?
#834 made an operator `enabled: false` on ticketing a hard kill switch across the organizer-facing UI: nav, ⌘K, all five ticket pages, the dashboard tile and the budget page. Adversarial QA confirmed each of those. It also found two organizer-visible outputs the deny does **not** reach.

## What still answers for a denied organization

**1. The tickets tRPC router** (`src/server/routers/tickets.ts`). `tickets.admin.getTicketTypes`, `getDiscountCodes`, `getDiscountCodesWithUsage`, `getPaymentDetails` and `create`/`deleteDiscountCode` are plain `adminProcedure` with no deny check. The UI can no longer reach them, but any authenticated organizer of that org can call them directly — and the discount endpoints still **write** to the tenant's own provider account.

This matters more than it looks because the platform is deliberately agent-facing: `konfctl` exists and an MCP server is planned, so "only reachable via the API" describes a growing surface rather than a shrinking one.

**2. The weekly Slack summary.** `buildTicketSection` (`src/lib/status/summary.ts`) reads live ticket data ungated, so a denied org keeps receiving live ticket counts in its weekly Slack post via the cron.

## The decision

Does the kill switch mean **the organizer-facing UI surface** — what #834 built, and what the module's own "NOT A SECURITY BOUNDARY" note implies — or **every organizer-visible output**?

Arguments for the current scope: the deny is a presentation control, and real revocation is removing credentials (which `resolveTicketingCredentials` already governs). Arguments for widening: an organizer who has been switched off should not keep getting Slack posts about ticket sales, and should not be able to mutate discount codes through an API the UI hides.

If widened, the natural shapes are a deny check in the tickets router (or shared middleware) and a gate on `buildTicketSection`.

## Not in scope either way

Deliberately ungated and correct: public ticket sales (attendee-facing, must not break mid-conference), workshop eligibility, and the status probe. Speaker-ticket issuance is borderline — the platform keeps writing discounts into a denied org's vendor account — but low-harm.

Comment corrected in #834 so the module no longer claims "no deep link outlives it".

Starefossen:
Implements the decided outcome of #836: an operator `enabled: false` on `ticketing` is a HARD kill switch that reaches **every organizer-visible output**, not only the UI #834 gated.

## What changed

**1. `requireFeatureNotDenied(featureId)` (`src/server/trpc.ts`)** — a new middleware, the deny-only half of `requireFeature`. It throws `FORBIDDEN` (`The "ticketing" feature has been switched off for this organization`) on an **active operator deny** and on nothing else, resolving through the existing registry (`isFeatureExplicitlyDeniedForOrg` → `featureOverrides`). No parallel mechanism.

**2. `tickets.admin.*` (`src/server/routers/tickets.ts`)** — the middleware is composed **once**, onto a `ticketingAdminProcedure` that all **13** procedures in the sub-router use, rather than six copy-pasted checks. Procedure fourteen inherits the gate by being declared there. That covers the endpoints named in the issue (`getTicketTypes`, `getDiscountCodes`, `getDiscountCodesWithUsage`, `getPaymentDetails`, `createDiscountCode`, `deleteDiscountCode`) plus the seven settings/page-content procedures the issue did not enumerate. A denied org now makes **no provider call and no Sanity write** — including the two mutations that wrote discount codes into the tenant's own provider account.

**3. `buildTicketSection` (`src/lib/status/summary.ts`)** — gated *before* `resolveTicketingProvider`, mirroring `resolveTicketingAdminAccess`'s order, so a denied org gets no ticket numbers in its weekly Slack post (or on the admin status page) and costs zero provider work. Sponsors and proposals are untouched: this switches off ticketing, not the weekly update. Added `isTicketingDeniedForConference` alongside the existing `isTicketingEnabledForConference`.

**4. Comment corrections** in `src/lib/features/ticketing.ts` and `src/lib/features/badges.ts` — both carried a now-false scope paragraph. They now state exactly what the deny reaches and what it does not.

## Why NOT `requireFeature('ticketing')` — and the hard constraint

`requireFeature` asks whether the org is **entitled**, which folds "never granted" together with "switched off". Verified rather than assumed: `organization-cloud-native-days` was created by `migrations/044-organization-tier-backfill` with **no `plan` and no `featureOverrides`**. `effectivePlan` normalizes an absent plan to `community`, and `ticketing` is `readiness: 'ga'` + `minPlan: 'pro'`, so `computeEntitlements` does **not** contain `ticketing` for that document — its ticketing works today via the platform-org grant and its own credentials, neither of which the entitlement set expresses. **An entitlement-shaped gate would have 403'd every ticketing call CNDN makes.** A deny-shaped gate cannot: CNDN carries no override, so `isFeatureExplicitlyDeniedForOrg` is `false` for it by construction.

That is not left to reasoning. Both new test files contain a **positive-control block** driven by the real CNDN document shape: all 13 procedures still answer, the provider is still reached for reads and writes, and the Slack summary still reports live counts — including under an expired deny, a deny on another feature, and a rejected Sanity read. Tighten the gate to entitlement semantics and those blocks fail.

## Sabotage proofs (full `pnpm test` per sabotage, baseline 504 files / 7344 tests on `origin/main`)

| Sabotage | Result |
| --- | --- |
| Drop `.use(requireFeatureNotDenied('ticketing'))` from `ticketingAdminProcedure` | **15 tests fail** in 1 file (`tickets.killswitch.test.ts`) — 13 procedure cases + the no-provider-call case + the paid-org case |
| Delete the deny early-return in `buildTicketSection` | **2 tests fail** in 1 file (`summary.ticketing.test.ts`) |
| No-op the middleware body in `requireFeatureNotDenied` | **18 tests fail** in 2 files (15 router + 3 middleware) |
| All restored | 507 files / 7390 tests pass |

## How I know these do not pass for the wrong reason

- **The router deny cases assert the refusal message verbatim, not just `FORBIDDEN`.** `adminProcedure`'s own waist also throws `FORBIDDEN` (`Admin privileges required`) — a code-only assertion is exactly the trap #842 hit. In these tests the caller **is** an organizer of the request org and the org resolves, so the waist admits the call; the only thing in the file that can produce `The "ticketing" feature has been switched off…` is the gate under test. The composition tests assert the *other* message explicitly, so the two refusals cannot be confused.
- **The Slack cases assert `resolveTicketingProvider` was never called.** `tickets: null` is also what an unconfigured or a failing conference produces, so that alone proves nothing. The fixture conference is fully configured with real sales, so without the gate the resolver runs and numbers come back — a summary in which the resolver was *not reached* can only come from the gate.
- **The gate runs for real, not stubbed.** Both files mock `getOrganizationById` and let `@/lib/features/platform-default` + `entitlements` execute, so override direction and expiry are genuinely exercised.

## Numbers (baseline = `origin/main`, after = this branch)

| Check | Before | After |
| --- | --- | --- |
| `pnpm test` | 504 files / 7344 tests passed | **507 files / 7390 tests passed** |
| `npx eslint .` | 0 errors, 201 warnings | **0 errors, 201 warnings** |
| `pnpm knip` | 1 config hint | 1 config hint (unchanged) |

## Deliberately left alone

- **Public/attendee ticket sales, workshop eligibility, the admin status probes** — out of scope per the issue; not gated, still working.
- **Speaker-ticket issuance** (`src/lib/events/handlers/speakerTicket.ts`) — **left ungated on purpose**, as the issue allows. It still mints a 100%-off coupon in a denied org's vendor account on speaker confirmation. It is borderline but low-harm: it is triggered by a proposal-confirmation event rather than by an organizer poking a ticketing surface, the coupon is deterministic per speaker email and idempotent, and gating it would silently break comp tickets for speakers of an org whose organizers were never told ticketing went off. Both feature modules' comments now name this hole rather than implying completeness.

### The one hole in the middleware itself, stated plainly

`requireFeatureNotDenied` does **not** fail closed on an unresolvable org: `isFeatureExplicitlyDeniedForOrg` reports a nullish org id, a missing document and a rejected Sanity read as "not denied", deliberately, so one flaky read cannot black out a working tenant (the hazard #828 fixed). Composed onto a procedure that does not resolve an org, an unresolvable request would pass this gate. It is closed **by composition** — `adminProcedure`'s waist already fails closed on an unresolvable org — and two tests assert that the waist, not this gate, is what refuses there. The doc comment says to compose it on `adminProcedure`, never on `publicProcedure`.

This gate is still **not a security boundary**: credential isolation remains `resolveTicketingCredentials` plus the tenancy guards. It is an operator's off switch, now honoured everywhere an organizer can see ticketing.

Closes #836

### DIFF
--- a/src/lib/features/badges.ts
+++ b/src/lib/features/badges.ts
@@ -63,9 +63,23 @@ import { isPlatformOrganization } from './platform'
  * either gate a security boundary: credential isolation is
  * `resolveTicketingCredentials` and the tenancy guards.
  *
- * The BADGES deny is still UI-only in the same sense — this module's own gate is
- * consumed by the badge pages and `issueBadgeForSpeaker`'s platform-org check;
- * no badge tRPC procedure composes the deny middleware today.
+ * The BADGES deny is still UI-only, and precisely so: every consumer of this
+ * module's gate is a rendering path — the `/admin/speakers` and
+ * `/admin/speakers/badge` pages, plus `./enabled.ts`, which feeds the admin
+ * layout's nav and ⌘K filtering. No badge tRPC procedure composes the deny
+ * middleware — `badge.admin.issue` is a plain `adminProcedure` — so an
+ * operator's deny does not reach the API at all.
+ *
+ * THE ISSUANCE TRIPWIRE IS NOT WHAT CLOSES THAT, and it is worth being exact
+ * about why: `src/lib/badge/issuance.ts` imports `getPlatformOrgId` from
+ * `@/lib/authz/platform` and imports NOTHING from this module, so it never
+ * consults this gate. It asks a different question — "is this the platform
+ * org?", not "did an operator switch badges off?" — and the answers coincide
+ * only in one direction. A denied NON-platform org is refused by the tripwire on
+ * the tripwire's own terms; a deny on the PLATFORM org stops at the pages, and
+ * `badge.admin.issue` still answers it. Closing that would mean composing
+ * `requireFeatureNotDenied('badges')` onto the badge router — deliberately out of
+ * scope for #836, which widened the TICKETING deny only.
  *
  * NO PLAN TIER, DELIBERATELY. `ticketing` moved to `minPlan: 'pro'` once the
  * capability became per-tenant (the customer's own provider account); badges
--- a/src/lib/features/badges.ts
+++ b/src/lib/features/badges.ts
@@ -51,17 +51,21 @@ import { isPlatformOrganization } from './platform'
  * not an absence, so it wins in both modules regardless of what the capability
  * says.
  *
- * SCOPE, SAID PRECISELY: the deny covers the ORGANIZER-FACING UI. It does not
- * reach the tRPC layer — `tickets.admin.*` (`getTicketTypes`,
- * `getDiscountCodes`, `getPaymentDetails`, `create`/`deleteDiscountCode`) are
- * plain `adminProcedure` and still answer for a denied org, and the discount
- * ones still WRITE to that tenant's own provider account. An API call is a deep
- * link; an earlier draft of this comment claimed none outlived the deny, which
- * was false. Nor does it reach the weekly Slack summary
- * (`buildTicketSection`), which keeps posting live ticket counts. Whether the
- * kill switch should mean every organizer-visible output rather than the UI
- * surface is an open owner decision — do not read this gate as a security
- * boundary in the meantime.
+ * SCOPE, SAID PRECISELY (owner decision on #836): a ticketing deny now covers
+ * every ORGANIZER-VISIBLE output, not just the UI. It reaches the organizer
+ * pages, the whole `tickets.admin.*` tRPC sub-router (via
+ * `requireFeatureNotDenied`, so the discount mutations no longer write to a
+ * switched-off tenant's provider account) and the weekly Slack summary's ticket
+ * section. It deliberately does NOT reach the ATTENDEE-facing paths (public
+ * ticket sales, workshop eligibility), the admin status probes, or
+ * speaker-ticket issuance — which still writes a discount code into a denied
+ * org's vendor account (borderline, low-harm, left knowingly). None of that makes
+ * either gate a security boundary: credential isolation is
+ * `resolveTicketingCredentials` and the tenancy guards.
+ *
+ * The BADGES deny is still UI-only in the same sense — this module's own gate is
+ * consumed by the badge pages and `issueBadgeForSpeaker`'s platform-org check;
+ * no badge tRPC procedure composes the deny middleware today.
  *
  * NO PLAN TIER, DELIBERATELY. `ticketing` moved to `minPlan: 'pro'` once the
  * capability became per-tenant (the customer's own provider account); badges
--- a/src/lib/features/ticketing.ts
+++ b/src/lib/features/ticketing.ts
@@ -55,10 +55,34 @@ import { isPlatformOrganization } from './platform'
  * absence of a decision. Swap those and you re-create the exact dead end #828
  * removed.
  *
- * NOT A SECURITY BOUNDARY, still. Credential isolation is enforced in
- * `resolveTicketingCredentials` and the tRPC tenancy guards; this decides what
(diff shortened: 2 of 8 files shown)


Starefossen:
Correction from production data — the PR's stated key finding is false.

Queried directly via the Sanity CLI (which works from this repo):

```
$ npx sanity documents query '*[_type=="organization"]{_id,name,"slug":slug.current,plan,featureOverrides}'
[
  {
    "_id": "organization-cloud-native-days",
    "featureOverrides": null,
    "name": "Cloud Native Days Norway",
    "plan": "pro",
    "slug": "cloud-native-days-norway"
  }
]
```

**CNDN carries `plan: "pro"`.** The PR asserts migration 044 left it with no plan, so it resolves to `community`, so `requireFeature('ticketing')` would have 403'd every CNDN ticketing call. Ticketing is `minPlan: 'pro'`, CNDN is pro, so it reaches minPlan and **would have been entitled by plan**. The plan was set to pro after 044 (owner decision, 2026-08-05: "Cloud Native Norway = PRO terms, not a comped override"). Reading the migration source and inferring current state is what went wrong.

This does not obviously make the PR incorrect — deny-only is still the right shape for a kill switch, and CNDN keeps access either way since `featureOverrides` is `null`. What it does affect:

1. **The comments are wrong in ~7 places** and state a specific falsehood about the production tenant — `tickets.killswitch.test.ts` (~156, 428-438, 614-619), `trpc.requireFeatureNotDenied.test.ts` (~847-882), and comment blocks at ~240 and ~363.
2. **The positive control may not guard what it claims to.** It is described as "driven by the real CNDN document shape", but it models a plan-less community org. If the real org is pro, that block would not necessarily fail if someone re-tightened the gate to `requireFeature` — which is precisely the regression it exists to catch. A guard calibrated to a document that does not exist is this project's canonical defect, not a cosmetic issue.
3. Minor: fixtures use slug `cloud-native-days`; production is `cloud-native-days-norway`.

Under adversarial review now with this evidence supplied. Not merging until the positive control is shown to fail when the gate is re-tightened.

PR #852, opened 2026-08-05
fix(gdpr): write speaker references weak so erasure is not blocked (#851)

### LINKED ISSUE #851: GDPR erasure is blocked again: strong speaker refs still being written after migration 041
Migration 041 weakened strong speaker references so a speaker document could be deleted. **It applied cleanly, its own verification passes, and the problem is back.** Production is accumulating new strong refs on the exact fields 041 existed to weaken.

## Verified in production (2026-08-06, read-only Sanity CLI)

```
count(*[_type=="notification" && defined(recipient._ref) && !(recipient._weak == true)])  ->  408
count(*[_type=="notification" && defined(actor._ref)     && !(actor._weak == true)])      ->  373
```

Newest such document: `_createdAt: 2026-08-05T04:00:12Z` — i.e. created **after** 041 ran on 2026-07-19, not left over from before it.

## Root cause

Schema `weak: true` governs **Studio writes and validation, not API writes.** `createReference()` does not emit `_weak`, so every server-side create writes a strong reference regardless of what the schema says.

The tell is in one object literal — `src/lib/notification/sanity.ts:72-92`:

```ts
recipient: createReference(item.recipientId),        // L72  — STRONG
...
doc.actor = createReference(item.actorId)            // L85  — STRONG
...
// Weak reference so a later proposal deletion doesn't orphan-block.
doc.relatedProposal = {
  ...createReference(item.relatedProposalId),
  _weak: true,                                       // L89-92 — correct
}
```

The author knew the pattern, wrote the comment, and applied it to the proposal ref while leaving the two **speaker** refs strong — which are precisely the ones 041 targets.

The correct pattern already exists elsewhere: the `assignedTo` path uses `{ ...createReference(id), _weak: true }` and holds up under the same check.

## Consequence

**A speaker who has received a notification since 2026-07-19 cannot be deleted.** Strong references block document deletion in Sanity, so a GDPR erasure request against any recently-active speaker fails today. This is a legal obligation, not a product nicety.

It also silently re-arms every time: 041 can be re-run, will report success, and the count will climb again from the next notification onward.

## Fix

Emit `_weak: true` on `recipient` and `actor` in `src/lib/notification/sanity.ts`, then re-run 041 to clean up the backlog. Worth auditing every other `createReference()` call site for the same omission rather than fixing these two in isolation — the schema gives no protection on the API path, so each site must carry it explicitly.

Consider whether `createReference()` should default to weak, with strong as the explicit opt-in. The current default makes the dangerous choice the quiet one.

## Note for whoever verifies 041 in future

A non-zero result from 041's verification query is **expected today** and does **not** mean the apply failed. Do not "fix" it by re-running the migration. Documented in #849.

Starefossen:
Fixes #851.

## What was wrong

`createReference()` returns `{ _type: 'reference', _ref }` — no `_weak`. Schema
`weak: true` governs Studio writes and validation, **not** API writes: reference
strength lives on the stored ref object. So every server-side write emitted a
STRONG reference regardless of the schema, and a strong reference blocks deletion
of its target — which is why a speaker who had received a notification since
migration 041 ran cannot be erased today.

## Measured in production (read-only Sanity CLI, 2026-08-05)

**Units matter here, so each row gives the command and its unit.** `count(*[…])`
returns **documents**, not references. A document can hold two offending
references (a notification with both `recipient` and `actor`), so the reference
total is larger than the document total — the two must not be conflated.

| field | `count(*[…])` → **documents** |
|---|---|
| `notification.recipient` strong | 408 |
| `notification.actor` strong | 373 |
| `message.author` strong | 13 |
| `conversation.createdBy` strong | 7 |
| `conversation.subjectSpeaker` strong | 1 |
| **sum of the five (references)** | **802 references** |
| **distinct documents affected** | **428 documents** |

The 428 figure is a single `count()` over the union predicate, not a sum. Newest
strong `notification.recipient`: `_createdAt 2026-08-05T04:00:12Z` — 041 ran
2026-07-19, so these are new writes, not a leftover backlog.

Note for whoever verifies the fix: **the Sanity CLI reports a zero count as an
error** (`Failed to run query: Query returned no results`), so wrap counts as
`{"n": count(*[…])}` — otherwise the success case of this fix reads as a failed
query.

## Scope: nine sites, not two

#851 names two. An audit of all 62 `createReference` / `createReferenceWithKey`
call sites found **nine** where the schema declares `weak: true` and the write
emits strong. All nine point at `speaker`; all nine are fixed. They include the
other three fields migration 041 exists to weaken, which the issue's repro did
not reach:

| file:line | doc.field | note |
|---|---|---|
| `src/lib/notification/sanity.ts:72` | `notification.recipient` | named in #851 |
| `src/lib/notification/sanity.ts:85` | `notification.actor` | named in #851 |
| `src/lib/notification/sanity.ts:267` | `notification.actor` | **patch `set`** |
| `src/lib/notification/sanity.ts:280` | `notification.recipient` | message-collapse `createIfNotExists` |
| `src/lib/messaging/sanity.ts:1225` | `conversation.createdBy` | proposal thread |
| `src/lib/messaging/sanity.ts:1284` | `conversation.createdBy` | sponsor thread |
| `src/lib/messaging/sanity.ts:1369` | `conversation.createdBy` | general thread |
| `src/lib/messaging/sanity.ts:1371` | `conversation.subjectSpeaker` | general thread |
| `src/lib/messaging/sanity.ts:1448` | `message.author` | every message sent |

`:267` matters beyond the backlog: a patch `set` **replaces** the stored ref
object, so on any active conversation the next message would re-strengthen a
document 041 had just weakened.

All nine use the existing house pattern — the shape `assignedTo` already used:
`{ ...createReference(id), _weak: true }`.

`conference` refs on the same documents stay strong on purpose: a conference is
not an erasure subject and referential integrity there is wanted.

## Migration 041 is still required after this merges

This stops the bleeding; it does not clean up the existing 802 strong references
across 428 documents. **Re-running 041 BEFORE this code fix would re-arm rather
than fix** — it would report success and the counts would climb again from the
next notification/message. Run it after this is deployed, via the Run Sanity
Migration workflow, as a separately authorised production step.

## Tests

This shipped because nothing asserted the shape of what gets written. Three
existing tests actively asserted the *buggy* shape — `party-model.test.ts`
asserted `authorParty.speaker` weak and `author` strong on adjacent lines, the
same split-brain the issue describes in the source. Corrected.

Added:

- **Behavioural** — asserting the document handed to `tx.create` /
  `tx.createIfNotExists` / the patch `set` payload, not a mock's return value,
  for every notification and messaging write path.
- **A repo-wide invariant** (`__tests__/lib/sanity/weak-references.test.ts`):
  parses `sanity/schemaTypes/*.ts` for every `(documentType, field)` declared
  `weak: true`, then parses every non-test file under `src/` and fails on any
  object literal that self-identifies as that `_type` and assigns the field a
  provably-strong reference. It also asserts the extracted schema map is
  non-empty and matches expectation, so it cannot pass vacuously.

  Its limits are documented in the file: it cannot see `.patch().set()` (no
  `_type` in the payload) and does not resolve refs built indirectly. It is a
  floor, not a proof — the patch paths are covered behaviourally instead.

Building the sabotage matrix also exposed a **coverage hole**:
`ensureSponsorConversation` had no test at all, so its `createdBy` could be
reverted to a strong ref with the full suite still green. Two tests added for it.

## Recommendation: do NOT invert `createReference()` to default weak

1. **It would not fix the class.** ~70 reference literals are hand-built
   (`{ _type: 'reference', _ref }`) and never touch the helper — including
   `speaker.organizations`, `sponsorActivity.createdBy` and
   `dashboardConfig.speaker`. Inverting buys a partial fix at full blast radius.
2. **The blast radius is unreviewable.** 52 of 62 call sites are strong today and
   many are strong *correctly* (`conference`, `tier`, `sponsor`, `conversation`).
   Flipping them changes delete-time integrity semantics repo-wide in a diff with
   no changed lines at the call sites.

To do it later, these would need to be true first: every schema field's intended
strength declared explicitly (many speaker-pointing fields are strong by
omission, not decision); an explicit `createStrongReference()` applied at the
intentional sites; the invariant test extended to hand-built literals and patch
payloads so the flip is verified rather than assumed; and a migration for
existing documents.

## Larger erasure surface — out of scope, needs its own issue

~17 further fields point at `speaker` and are strong in **both** schema and code:
`talk.speakers`, `conference.organizers`, `conference.teams[].members`,
`imageGallery.speakers`/`untaggedSpeakers`, `review.reviewer`,
`coSpeakerInvitation.invitedBy`/`acceptedSpeaker`, `sponsorActivity.createdBy`,
`sponsorForConference.assignedTo`, `travelSupport.speaker`/`reviewedBy`,
`volunteer.reviewedBy`, `invitationLetter.issuedBy`, `speakerBadge.speaker`,
`dashboardConfig.speaker`, `conversationPreference.speaker`, `schedule.owner`.
Each also blocks erasure. Unlike the nine above these are not code contradicting
the schema — the schema says strong — so changing them is a schema/product
decision plus migrations, not a bug fix.

Two further audit findings, reported rather than guessed at:
`src/lib/schedule/sanity.ts:411` writes a *speaker* id into `schedule.owner`,
which the schema declares `to: [{ type: 'staff' }]`; and
`src/app/api/cron/contract-reminders/route.ts:146` writes
`createdBy: { _ref: 'system' }` — a strong reference to an id with no document.

## Risk

Low, and partly pre-proven: 698 `notification` **documents** already hold a weak
`recipient` (041 weakened them 2026-07-19) and the inbox has read them fine
since. Weak refs deref identically; the only behavioural difference is that
`actor->` / `author->` may resolve null once a target is erased — the intended
outcome, and what the schema already promised.

### DIFF
--- a/__tests__/lib/messaging/party-model.test.ts
+++ b/__tests__/lib/messaging/party-model.test.ts
@@ -45,6 +45,7 @@ import {
   resolveRecipients,
   canAccessConversation,
   ensureProposalConversation,
+  ensureSponsorConversation,
   createGeneralConversation,
   addMessage,
   syncProposalConversationParticipants,
@@ -402,7 +403,11 @@ describe('dual-write — ensureProposalConversation writes participants[] alongs
 
     const doc = tx.createIfNotExists.mock.calls[0][0] as Record<string, unknown>
     // Legacy write-source fields are UNCHANGED (still present).
-    expect(doc.createdBy).toEqual({ _type: 'reference', _ref: 'sp-1' })
+    expect(doc.createdBy).toEqual({
+      _type: 'reference',
+      _ref: 'sp-1',
+      _weak: true,
+    })
     expect(doc.proposal).toEqual({
       _type: 'reference',
       _ref: 'prop-1',
@@ -439,7 +444,11 @@ describe('dual-write — createGeneralConversation', () => {
     })
     const doc = writeMock.create.mock.calls[0][0] as Record<string, unknown>
     // Legacy fields unchanged: creator ref present, no subjectSpeaker.
-    expect(doc.createdBy).toEqual({ _type: 'reference', _ref: 'sp-9' })
+    expect(doc.createdBy).toEqual({
+      _type: 'reference',
+      _ref: 'sp-9',
+      _weak: true,
+    })
     expect('subjectSpeaker' in doc).toBe(false)
     expect(doc.participants).toEqual([
       storedSpeaker('sp-9'),
@@ -455,7 +464,11 @@ describe('dual-write — createGeneralConversation', () => {
       subjectSpeakerId: 'sp-7',
     })
     const doc = writeMock.create.mock.calls[0][0] as Record<string, unknown>
-    expect(doc.subjectSpeaker).toEqual({ _type: 'reference', _ref: 'sp-7' })
+    expect(doc.subjectSpeaker).toEqual({
+      _type: 'reference',
+      _ref: 'sp-7',
+      _weak: true,
+    })
     expect(doc.participants).toEqual([
       storedSpeaker('org-1'),
       storedSpeaker('sp-7'),
@@ -475,8 +488,12 @@ describe('dual-write — addMessage writes authorParty alongside the author ref'
     })
 
     const doc = tx.create.mock.calls[0][0] as Record<string, unknown>
-    // Legacy author ref unchanged.
-    expect(doc.author).toEqual({ _type: 'reference', _ref: 'sp-9' })
+    // Legacy author ref — WEAK, like the party form beside it (#851).
+    expect(doc.author).toEqual({
+      _type: 'reference',
+      _ref: 'sp-9',
+      _weak: true,
+    })
     // Dual-written party form of the author.
     expect(doc.authorParty).toEqual({
       partyType: 'speaker',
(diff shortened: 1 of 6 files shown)


Starefossen:
## Sabotage proof

Each of the nine `_weak: true` values was reverted to the strong form **one at a
time**, with the **full** suite re-run each time and restored afterwards.

The load-bearing check is the **total** in each row: it stays `(508)` files and
`(7423)` tests in every single case. That is the guard against a malformed edit
breaking an import, silently dropping a test file, and reading as a pass — the
failure count rises while the total never moves, so every test still ran.

| # | reverted site | Test Files | Tests | failing file(s) |
|---|---|---|---|---|
| N1 | `notification.recipient` (create) | 2 failed \| 506 passed **(508)** | 4 failed \| 7419 passed **(7423)** | notification, **weak-references** |
| N2 | `notification.actor` (create) | 1 failed \| 507 passed **(508)** | 3 failed \| 7420 passed **(7423)** | notification |
| N3 | `notification.actor` (patch `set`) | 1 failed \| 507 passed **(508)** | 3 failed \| 7420 passed **(7423)** | notification |
| N4 | `notification.recipient` (upsert) | 2 failed \| 506 passed **(508)** | 4 failed \| 7419 passed **(7423)** | notification, **weak-references** |
| M1 | `conversation.createdBy` (proposal) | 2 failed \| 506 passed **(508)** | 3 failed \| 7420 passed **(7423)** | party-model, **weak-references** |
| M2 | `conversation.createdBy` (sponsor) | 1 failed \| 507 passed **(508)** | 1 failed \| 7422 passed **(7423)** | party-model |
| M3 | `conversation.createdBy` (general) | 2 failed \| 506 passed **(508)** | 3 failed \| 7420 passed **(7423)** | party-model, **weak-references** |
| M4 | `conversation.subjectSpeaker` | 2 failed \| 506 passed **(508)** | 3 failed \| 7420 passed **(7423)** | party-model, messaging/sanity |
| M5 | `message.author` | 2 failed \| 506 passed **(508)** | 3 failed \| 7420 passed **(7423)** | party-model, messaging/sanity |
| — | **restored (control)** | **508 passed (508)** | **7423 passed (7423)** | none, exit 0 |

Every case exits non-zero. Named assertions that fired, e.g. N3:

```
x upsertMessageNotifications: actor is weak in the patch `set` —
  a strong value here would RE-STRENGTHEN an already-weakened document
x no notification write path emits a strong speaker ref
x NEW: seeds count 1 via createIfNotExists and patches the same values
```

### Two things the matrix proves about the tests themselves

**M2 fails on exactly one test — the one added in this PR.** Before it,
`ensureSponsorConversation` had no coverage at all, so reverting that
`createdBy` to a strong ref left the full suite green. That is the same failure
mode as #851 itself: an unasserted write path drifting from the schema.

**The invariant test's documented blind spots are real and visible here.**
`weak-references.test.ts` fires on N1, N4, M1 and M3 but *not* on N3, M4 or M5 —
N3 is a `.patch().set()` payload (no `_type` to key on), and M4/M5 sit in
conditional/spread literals that carry no `_type`. This is the limitation stated
in the test's own header, confirmed rather than assumed, and it is why those
three paths are covered behaviourally instead. The invariant test is a floor
against future call sites, not a proof of reference strength in general.

### Before / after

Baseline is `origin/main` @ bacbbe7c.

| check | before | after |
|---|---|---|
| `pnpm test` | 507 files / 7412 tests, all pass | **508 files / 7423 tests, all pass** |
| `npx eslint .` | 0 errors, 201 warnings | 0 errors, 201 warnings |

(`npx eslint .`, not `next lint` — the latter is vacuous in Next 16. The 201
warnings are the pre-existing `tenancy/no-unscoped-groq` set, unchanged.)

PR #874, opened 2026-08-07
Fix the I'm speaking card on the speaker dashboard

schildwaechter:
Changed the speaker dashboard so the I'm speaking at card reads the current speaker document instead of the session snapshot. 

What changed:
- Added getSpeaker lookup for the signed-in speaker.
- Built speakerWithTalks from currentSpeaker ?? session.speaker.

Effect:
- The share card now uses the up-to-date saved profile image from Sanity.
- If the live lookup ever fails, it falls back to the old session behavior.

Note: GPT-5.4 and as websites of development tooling `mise` and `fnox` are gone, can't test anything locally.


schildwaechter:
Too little understanding for the problem, opened https://github.com/CloudNativeBergen/website/issues/875 instead.

Issue #808, opened 2026-08-05
SECURITY: speaker merge injects an attacker-chosen address into the victim's VERIFIED match-set


Starefossen:
Found while fixing #742. Same class, **strictly worse**: #742 let an organizer point a victim's *unverified* display `email` at an address they control. This one writes the attacker's address into the victim's **`knownEmails`** — the set the whole login path treats as verified-owned — so it survives every narrowing being considered in #807.

Not introduced by #731 or #740; both merge arms have looked like this since #449.

## The chain

Every step is an ordinary organizer of tenant A using the app. No Sanity access.

1. **Create a throwaway.** `speaker.admin.create` takes an arbitrary `email` and stamps `organizations: [ORG_A]` (`src/server/routers/speaker.ts`). Attacker creates speaker **L** with `email: attacker@evil.example`. L is exclusive to A and referenced by nothing.
2. **Pick a victim.** Any speaker **V** the ordinary guard admits — membership *or* participation in A. `ensureSpeakerOrgMembership` stamps the current org on **every login**, so that is anyone who has ever signed into A's site, including an organizer of tenant B.
3. **Merge L into V.** `speaker.admin.merge({ survivorId: V, loserId: L })`. The loser arm is `requireExclusive` and L passes it. The **survivor arm is ordinary standing** — `await requireSpeakerInCurrentOrg(input.survivorId)` — so V passes.
4. `computeSurvivorFieldMerge` (`src/lib/speaker/merge.ts`) unions **both display emails** into the survivor's match-set:

   ```ts
   const knownAfter = uniqueEmails([
     ...(survivor.knownEmails ?? []),
     ...(loser.knownEmails ?? []),
     survivor.email,
     loser.email,          // ← attacker@evil.example
   ])
   ```

   and `mergeSpeakers` commits it: `transaction.patch(survivorId, (p) => p.set(plan.survivorSet))`.
5. **Sign in.** Attacker authenticates as `attacker@evil.example` — OAuth or the #740 email link. `findSpeakersByEmails` matches V on `knownEmails`, `linkAndAccrue` welds the attacker's provider account onto V's document, and the session carries V's `organizerOrgIds`. If V organizes tenant B, the attacker is now an organizer of B.

`mergePreview` takes the same two arms and would show the operation as permitted.

## Why the existing controls miss it

- `adminProcedure` gates the **caller**, not the survivor.
[...]

PR #880, opened 2026-08-08
feat(admin): invite a co-organizer by email, proved by magic-link sign-in

Starefossen:
Closes the website half of **RunKonf/platform#49** (phases 1–2) — the substance of **M2** in RunKonf/platform#57.

## The gap

`conference.organizers[]` is the canonical `/admin` grant. Its only editor was a picker whose corpus is *this conference's confirmed speakers plus its current organizers* — so on a fresh tenant the corpus is **exactly the founder**, and there is nobody to pick. Every addition has been a hand-edit in Sanity.

kontroll structurally cannot build this: `conference` is off its write partition, and its membership keys on `userKey` while organizer standing keys on `speaker._id` — disjoint identity spaces with no join beyond heuristic email matching. So this is ours.

## What ships

Invite a person to `conference.organizers[]` **by email address**, accepted by proving control of that mailbox with an email magic-link sign-in.

| Piece | Where |
| --- | --- |
| `organizerInvitation` document type | `sanity/schemaTypes/organizerInvitation.ts` |
| Signed token (mint + **verify**) | `src/lib/organizer-invite/token.ts` |
| Conference-scoped reads + retention purge | `src/lib/organizer-invite/sanity.ts` |
| `list` / `invite` / `revoke` / `accept` | `src/server/routers/organizerInvite.ts` |
| The redeemed address, as a session fact | `src/lib/auth/email-link/identity.ts` |
| Accept page + panel (6 states) | `src/app/(main)/organizer-invitation/accept/`, `src/components/organizer-invite/` |
| Admin "invite by email" modal | `src/components/organizer-invite/OrganizerInvitesEditor.tsx`, mounted in `/admin/settings` |

Nothing was rebuilt: it reuses the shipped email magic-link sign-in, `getOrCreateSpeakerForVerifiedEmail`, the generic Resend `sendEmail` wrapper, `requireDocumentInCurrentOrg`, `scopedFetch`, and the existing daily cleanup cron.

**A standalone document, not a field on `conference`.** `conference` is append-only and read straight from Sanity by kontroll; keeping a high-churn, short-lived list off it avoids putting invitation traffic on the cross-app contract. `sanity/schema-shape.baseline.json` is untouched.

## The security core — and the thing I got wrong first

**The invitation token is not ownership proof.** Invitation mail is forwarded and lands in shared inboxes.

My first implementation read `speaker.providers[]` for an `email-link:<invited address>` entry. Only `getOrCreateSpeakerForVerifiedEmail` *mints* one, and only after a real redemption — so it looked like proof. **Adversarial review showed it is launderable.** `mergeSpeakers` UNIONS a loser's `providers[]` onto the survivor, so an organizer who can arrange to merge a document carrying `email-link:victim@x` moves that entry onto their own and can present it as proof of a mailbox they never controlled. Separately, it only ever proved *"controlled it at some point"*, not *"in this session"*.

`providers[]` is an accumulating dedup key. It was never an assertion about the person holding the session, and an admin grant should not key on one.

**The fix: mint the fact where it is actually true.** `authorize` now carries the verified address onto the JWT at the moment of redemption, and `accept` reads `session.emailLinkIdentifier`. This is a **scope extension into `src/lib/auth.ts` and deserves a look** — it is deliberately additive: one new optional claim that nothing else reads, absent wherever it was previously absent, so every consumer fails closed. OAuth accept (phase 3, gated on #808) stays deferred *by construction* rather than by a separate check. No existing behaviour changes; the cost to an old session is one more sign-in, which is the action being asked for anyway.

**Acceptance never writes identity.** The only writes are `conference.organizers[]` and the invitation's own status — a test asserts the speaker document is never touched.

### Check order in `accept` — load-bearing

```
token signature → conference-scoped lookup → stored-token compare
→ OWNERSHIP → expiry → status → grant
```

Ownership is now **first**. Review caught that the status gate sat above it and threw a distinct `BAD_REQUEST`, so a forwarded-token holder could distinguish pending from revoked/accepted — an existence-and-state oracle that undid the whole point of ownership-before-expiry. The accept page had the same inversion and leaked more than the mutation; both are fixed and now share one order. Everything above ownership refuses with one identical `NOT_FOUND`, and the three ownership branches share one message (a test asserts the message set has size 1). Status sits below expiry so a lapsed invitation gets the expiry message rather than a generic one.

### Two hardenings over the co-speaker token

1. **It is actually verified.** `createInvitationToken` signs and then never checks — redemption is an exact string lookup, so its signature participates in no decision. Here a forged token is refused **before any Sanity read**.
2. **Domain separation + `timingSafeEqual`.** The HMAC covers a `konf.organizer-invite.v1` label, so a token minted for another purpose under the same secret cannot be replayed.

## Sabotage matrix

Each guard removed, the suite re-run, then restored (final state: 0 failed, 132 passed across the feature's five suites).

| Guard removed | Tests that caught it |
| --- | --- |
| `adminProcedure` → `organizerProcedure` on `invite`/`revoke`/`list` | 3 — the CROSS-ORG trio |
| `accept` requires the session's proof to match the invited address | 8 |
| Ownership block swapped **after** expiry | 1 — "…triggers NO expiry write" |
| Status gate moved back **above** ownership | 1 — "NO LIFECYCLE ORACLE" |
| `auth.ts` stamps the redeemed address onto the JWT | 1 |
| `sessionCallback` passes the claim through | 1 |
| `emailLinkIdentifierOf` rejects a blank/non-string claim | 2 |
| The claim is stamped **only** in the email-link branch | 1 |
| Grant transaction's `ifRevisionId` | 1 |
| `requireDocumentInCurrentOrg` in `revoke` | 3 — incl. type confusion (a `review` id) |
| Token signature verification | 1 — "GUARD BEFORE FETCH" |
| HMAC domain-separation label | 5 |
| Presented-vs-stored token compare | 1 |
| Conference scope on the by-id read | 2 |
| Conference scope on the list read | 1 |
| Nested conference filter in `isEmailAlreadyOrganizer` | 1 |
| Purge keeps ACCEPTED invitations | 1 |
| Rollback of a token-less invitation | 1 |
| Purge restricted to `_type == "organizerInvitation"` | 1 |
| Purge is global, not conference-scoped | 2 |
| `accept` refuses an impersonating session | 1 |
| `accept` refuses a canonical/normalized disagreement | 1 |
| `revoke` is revision-conditioned | 1 |
| `isOrganizerInvitationExpired` stays narrow | 1 |
| The accept **page** decides ownership before lifecycle | 1 |
| The claim stamp validates before writing | 4 |

Two process notes, because both are the failure mode AGENTS.md warns about:

- My first attempt at row 1 swapped the identifier **without importing it**, breaking module load. The suite reported "no tests" and my counting script read that as a pass. Redone properly it fails on the action succeeding.
- The four auth-wiring rows initially showed **zero** failures — the router tests construct a session object directly, so nothing exercised `auth.ts` at all. That gap is now covered in `__tests__/lib/auth/email-link-identity.test.ts`.

A **second adversarial pass** then mutation-tested the rework and found four more things the tests did not hold, all now fixed and in the table above. The worst was the purge: removing `_type == "organizerInvitation"` from an unscoped, cross-tenant `clientWrite.delete({query})` **survived the entire suite** — and `coSpeakerInvitation` and `provisioningRequest` both carry `status` + `expiresAt`, so that clause is all that stands between this cron and deleting every tenant's lapsed ones nightly. Fixtures of both types are now in the dataset.

## What is executed rather than asserted

- **`__tests__/lib/organizer-invite/sanity.test.ts`** stubs only the *transport*: `clientReadUncached.fetch` is a real `groq-js` evaluator over a two-tenant dataset, and `scopedFetch` is **not** mocked — so the query text under test is the text that runs, prepended predicate included. Review's sharpest finding was that this file had **no tests at all** while every tenancy claim lives in it, and the router's "CROSS-TENANT" test was `mockResolvedValue(null)` — a statement about the test, not the query. Three scope sabotages are now caught, each failing on a foreign document *resolving*.
- **`grant-scope.test.ts`** reads `ORGANIZER_ORG_IDS_FIELD` out of `src/lib/speaker/sanity.ts` **as text** and evaluates it, proving a grant yields `['org-a']` and never `org-b`, and pinning that the deprecated global `isOrganizer` flag *does* flip everywhere — which is why nothing may gate on it. It pins the literal, not its use; stated in the file.
- `node:crypto` runs for real in the token tests, and `ifRevisionId` was verified against the real `@sanity/client` (it serializes to `ifRevisionID`, and `transaction().patch()` accepts a built Patch) rather than assumed from a mock.

## Numbers

| | before | after |
| --- | --- | --- |
| Test files | 524 | **530** |
| Tests | 7705 | **7825** |
| `tenancy/no-unscoped-groq` | 200 warnings / 51 files (baseline 201/51) | **200 / 51 — unchanged** |

New files contribute **zero** tenancy warnings and do not appear in `no-unscoped-groq.baseline.json`; the ratchet was not raised. `pnpm run check` green; `pnpm build-storybook` succeeds.

## Privacy

`/privacy` gains an **Organizer Invitations** section: we now hold an email address belonging to someone who may never have used the site. Review caught the first draft asserting three things the code did not do — that revoking *removes the record* (it only set a status), that *accepting creates or links a profile* (sign-in does that; accept never touches `speaker`), and that withdrawn invitations are deleted *promptly* (a withdrawn invitation is retained until its original expiry). All corrected, and a **retention purge** was added to the daily cron so the erasure statement is true rather than aspirational: non-accepted invitations are deleted once past `expiresAt`; accepted ones are kept as the provenance of a live admin grant, and the cron now reports the count alongside its other passes.

## Holes, named

- **Local `pnpm run build` fails** with `Failed to collect page data for /api/cron/contract-reminders`. I built `origin/main` in a clean worktree and got the identical failure — local environment, not this branch. **CI's Build and Runtime Smoke Test both pass.**
- **No visual verification.** Playwright is broken here, so `pnpm shoot` cannot run. Twelve stories instead (seven accept-panel states, five invites-editor including empty, loading and load-failed) — CI Storybook is the evidence.
- **A non-owner still learns two things** from the accept page: that an invitation exists, and its address masked to first-initial + domain. Both are unavoidable if a real invitee signed in with the wrong identity is to be told which mailbox to use, and the mutation concedes the same existence bit by answering `FORBIDDEN` rather than `NOT_FOUND`. Stated as a deliberate trade in the page's own docblock.
- **The bearer token is stored in cleartext** on a Studio-visible document (as `coSpeakerInvitation` is) and is not cleared after accept. It is no longer sufficient for anything on its own — the proof is a session fact — but it is stated rather than implied.
- **Concurrency is guarded at the invitation, not the conference.** Both `accept` and `revoke` are revision-conditioned on the invitation (the second pass caught that only `accept` was, so the race was closed in one direction). The `organizers[]` append itself is unguarded, so an accept racing `conference.updateOrganizers` can duplicate a `_ref` — harmless for authz (`organizerOrgIds` dedupes) but it renders a duplicate row.
- **Impersonation is refused by `accept`.** It swaps `session.speaker` while leaving the proof intact, which would decouple the mailbox proved from the speaker granted. Production strips the parameter anyway; the guard is there because the coupling is an invariant of the grant, not of the environment.
- **A forwarded token remains a permanent existence oracle** for the document's lifetime — ownership precedes status, so even a long-revoked invitation answers `FORBIDDEN` rather than `NOT_FOUND`. Bounded by the retention purge.
- **No `decline`** for an invitee (the co-speaker module has one), and **no notification** to other organizers on issue/accept. Both deliberate scope trims.
- **Revocation here revokes an INVITATION.** Removing a sitting organizer stays with `conference.updateOrganizers`, whose self-lockout guard already makes the schema's `min(1)` unreachable. `accept` only ever **appends** — a test asserts no full-array `set` on `organizers` exists in this path.

## Not in scope

OAuth-session acceptance (website#808), reviewer roles (platform#30), anything in kontroll.

**kontroll copy:** its org page says *"someone already on it adds you from that conference's admin settings."* That describes a door that does not exist today — **once this ships it becomes true.** Not fixed here (different repo).

## Docs

`docs/AUTH.md` gains the acceptance-proof section, including why the `providers[]` approach was wrong; `docs/CONTROL_MODEL.md` gains the `organizerInvitation` write-partition row and why this write cannot live in kontroll. Both current-state, not history.

Refs RunKonf/platform#49, RunKonf/platform#57.


Starefossen:
## Adversarial review (independent pass)

Re-verified the judgement calls rather than the fixed findings. Sabotages re-run locally from a clean worktree at `474259d9`; full suite 530 files / 7825 tests green, matching the claimed counts exactly.

### Confirmed
- **Auth scope extension is sound and genuinely additive.** The laundering rationale is real: `src/lib/speaker/merge.ts` unions a loser's `providers[]` onto the survivor (pinned by `merge.test.ts`), so document state can carry a proof the session holder never earned. The claim is minted only in the email-link `authorize` → email-link branch of `jwtSignInCallback`; `applySpeakerToToken` sets explicit fields only, no spread; the OAuth-smuggle case (user object carrying the field) is pinned at `email-link-identity.test.ts:365`; `sessionCallback` only adds the key when present; the sole consumers are `organizerInvite.accept` and the accept page, both failing closed on absence. Removing the JWT stamp fails 2 tests.
- **The purge type clause is now load-bearing in a test.** Dropping `_type == "organizerInvitation"` fails `deletes ONLY organizerInvitation documents` against real groq-js evaluation with `coSpeakerInvitation`/`provisioningRequest` fixtures. The other two predicates are pinned too (accepted-kept, live-pending-kept), and `expiresAt < $now` fails closed on a missing field (GROQ null comparison is falsy).
- **Ordering** — moving the status gate back above ownership fails `NO LIFECYCLE ORACLE`; the page's parametrized `same state for %s` test pins the page order non-vacuously (owner-sees-precise-state companions included).
- **Race** — both `accept` and `revoke` are revision-conditioned on the invitation; the accept transaction guards the whole commit.
- **Tenancy** — `sanity.test.ts` evaluates the real scope-prepended queries over a two-tenant dataset with an adversarial same-address foreign invitation; `scopedFetch` throws on empty scope, so an unresolvable tenant cannot fail open.

### Findings (all non-blocking)
1. **Stale security contract in `__tests__/api/trpc/organizer-invite.test.ts:16-19`** — the header still names the rejected `providers[]` mechanism as THE PROOF. Copilot flagged this (suppressed comment) and it was not fixed. It contradicts the file's own tests; fix the comment.
2. **The organizers[] race has a worse direction than the stated one.** `conference.updateOrganizers` does an unguarded full `set` (`conference.ts:570`), so a save racing an accept can silently WIPE the just-granted organizer — and the invitation is already `accepted`, so the door is spent; recovery is a fresh invitation. The PR body names only the duplicate-`_ref` direction. Still non-blocking (tiny window, recoverable, pre-existing pattern), but the stated hole is understated — worth a line in the docblock.
3. The expiry-marking write in `accept` is the one unconditional patch left: racing a `revoke` it can overwrite `revoked` with `expired`. Both terminal and purge-eligible, so consequence ~nil; note only.
4. `OrganizerInviteAcceptSchema.token` / `RevokeSchema.invitationId` have no `.max()` (email is capped at 254): an oversized token reaches base64+HMAC before refusal. Bounded by transport limits; cheap hardening.
5. "Bounded by the retention purge" for the existence oracle holds only for non-accepted invitations — accepted ones are purge-exempt, so their forwarded token answers FORBIDDEN forever. Wording nit.

### Bot threads
Zero unresolved review threads. Copilot's four suppressed comments: page ordering, router status ordering and privacy wording were all fixed in later commits; the test-header one is finding 1 above. Qodo is billing-paused — nothing to rule on. Chromatic `UI Tests` pending = quota, non-required.

**VERDICT: NON-BLOCKING**

